PR #9, opened 2016-04-20
Add aggregate point in polygon analysis

dgaubert:
cc @rochoa


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "camshaft",
-  "version": "0.4.0",
+  "version": "0.5.1",
```
rochoa: I would not increment versions in PR, but at release time.

reference/versions/0.5.0/reference.json
```diff
@@ -0,0 +1,112 @@
+{
```
rochoa: Same as with the version bump, I would do this once PR is merged into master and we want to release it.

lib/node/nodes/point-in-polygon.js
```diff
@@ -1,45 +0,0 @@
-'use strict';
```
rochoa: We should not remove existing analyses, otherwise we will break some applications.

I know this is not 1.x version but I prefer to keep them all for now.

rochoa:
Can you take a look at my comments and we do a second CR 😄 .

reference/versions/0.5.0/reference.json
dgaubert: So, I can't add new analysis without updating references or without testing it. How can I proceed to implement new analysis?.

dgaubert: I understand that a new analysis requires two PR, first one to implement the node and test it isolated and second one to integrate it into analysis workflow. right?

rochoa: That's my fault: I started using `AnalysisGraph` from reference within workflow/factory but I shouldn't have. We need an alternative within `./lib` to `getChildNodesNames` from a definition so we don't need to use anything from `./reference`.

dgaubert: So `./reference` should be for camshaft's dependants and `./lib` should introspect to discover its own analysis, right?

rochoa: https://github.com/CartoDB/camshaft/pull/11 should do the work 😅

dgaubert: thx!

rochoa:
🚢 🇮🇹

PR #10, opened 2016-04-22
Analysis trade area

dgaubert:
cc @rochoa


package.json
```diff
@@ -1,6 +1,6 @@
 {
```
rochoa: Do not bump version neither generate reference in PR.

rochoa:
We are missing the `disolved` option.

lib/node/nodes/trade-area/trade-area-query-builder.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
```
rochoa: I was thinking more about a generic method to load any node/*.sql file as a `dot.template`.

lib/node/nodes/trade-area/trade-area.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
```
rochoa: I prefer to keep the nodes inside `lib/node/nodes` and not create new directories.

lib/node/nodes/trade-area/trade-area.js
```diff
@@ -0,0 +1,34 @@
...
+
+var Node = require('../../node');
+var buildQuery = require('./trade-area-query-builder');
+var buildRange = require('./data-range-builder');
```
rochoa: I would prefer to keep it everything in here: no builders, etc.

If a node is too complex probably we have to consider splitting it into different nodes.

lib/node/node.js
```diff
@@ -212,6 +214,14 @@ module.exports.getChildNodesNames = function(NodeClass) {
     }, []);
 };
 
+module.exports.getTemplate = function (name) {
+    var template = dot.template(fs.readFileSync(__dirname + '/nodes/sql/' + name +'.sql'));
+
+    return function (it) {
```
rochoa: This is the same as `return dot.template(...)`, isn't it?

rochoa:
GO, GO, GO!

dgaubert:
I've just applied changes from `nodes-refactor-tplfn`

lib/node/node.js
dgaubert: Ups, I don't know what I was thinking.

PR #15, opened 2016-05-03
Nodes refactor template function

rochoa:
@dgaubert I was thinking about a much more simple way to load template functions so we don't need to create intermediate builders. What do you think?


dgaubert:
totally agree. I'm not aware about how nodes are going to evolve. I'm just trying to apply SRP.

dgaubert:
I guess this PR is just for showing the path to accomplish that, right?

rochoa:
While I understand your point about SRP I prefer to keep it simple so anyone is able to look at a file and know how to create another type of node.

Yes, this PR was just to validate we can change our approach.

rochoa:
Oh, my intention was to merge this into your branch but I guess it no longer makes sense, right? Should I close it?

dgaubert:
I think so

On Tue, May 3, 2016 at 2:33 PM, Raul Ochoa <notifications@github.com> wrote:

> Oh, my intention was to merge this into your branch but I guess it no
> longer makes sense, right? Should I close it?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/CartoDB/camshaft/pull/15#issuecomment-216513733>
>



-- 

*Daniel García Aubert*
Developer

*CartoDB*
Plaza de Callao 4, 2 28013 Madrid
www.cartodb.com

PR #18, opened 2016-05-11
Allow to have optional/nullable params

rochoa:
You can define a param like:

```js
Node.PARAM.NULLABLE(Node.PARAM.STRING)
```

cc @dgaubert

### DIFF
--- a/lib/node/node.js
+++ b/lib/node/node.js
@@ -158,10 +158,11 @@ var TYPE = {
     ENUM: 'enum'
 };
 
-function Validator(type, isValid) {
+function Validator(type, isValid, nullable) {
     this.type = type;
     this.isValid = isValid;
     this.expects = type;
+    this.nullable = nullable || false;
 }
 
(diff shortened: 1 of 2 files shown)


lib/node/node.js
```diff
@@ -228,13 +234,13 @@ module.exports.create = function createNode(type, expectedParams, options) {
...
         this.json.type = this.type;
         Object.keys(expectedParams).forEach(function(expectedParamName) {
-            if (!this.params.hasOwnProperty(expectedParamName)) {
-                throw new Error('Missing required param "' + expectedParamName + '"');
-            }
-
             var param = this.params[expectedParamName];
             var validator = expectedParams[expectedParamName];
 
+            if (!validator.nullable && !this.params.hasOwnProperty(expectedParamName)) {
```
dgaubert: Oh!. That was the place. When I tried it, I put this check at the wrong place and I did that the implementation was awkward.

dgaubert:
👍

PR #27, opened 2016-05-25
Integrates fallback queries

### LINKED ISSUE #25: Use fallback queries from Batch API
So we setup node status with onerror/onsuccess callback queries instead of using an array of queries that might fail.

dgaubert:
Fixes #25, Used Batch API fallback  jobs in order to persist node analysis.

It's working and tests are in green but I have some doubts about the implementation. Please, see inline comments.

cc @rochoa


lib/service/database.js
```diff
@@ -12,6 +12,7 @@ var debug = require('../util/debug')('database-service');
 var QueryRunner = require('../postgresql/query-runner');
 var QueryParser = require('../postgresql/query-parser');
 var BatchClient = require('../postgresql/batch-client');
+var Job = require('./job');
```
dgaubert: encapsulates some ugly details about fallback-jobs

lib/service/database.js
```diff
@@ -276,22 +277,37 @@ DatabaseService.prototype.queueAnalysisOperations = function(analysis, callback)
...
             nodeToUpdate.setStatus(status);
         });
 
         nodesToQueueForUpdate.forEach(function(nodeToUpdate) {
             if (nodeToUpdate.shouldCacheQuery()) {
                 var targetTableName = nodeToUpdate.getTargetTable();
-                asyncQueries.push(deleteFromCacheTableQuery(targetTableName));
-                asyncQueries.push(updateNodeAtAnalysisCatalogQuery(nodeToUpdate.id(), Node.STATUS.RUNNING));
-                asyncQueries.push(populateCacheTableQuery(targetTableName, nodeToUpdate.sql()));
+                asyncQueries.push({
```
dgaubert: duplicated code, but It's needed due to the loop goes through several nodes and we need to pass the specific node-id to the fallback.

I'd like explore other ways to do that.

lib/postgresql/batch-client.js
```diff
@@ -50,16 +50,45 @@ function runQueries(queryRunner, queries, callback) {
     }
 
     async.eachSeries(
-        queries,
+        queries.query,
         function(query, done) {
-            queryRunner.run(query, false, done);
+            if (typeof query === 'string') {
```
dgaubert: Definitely this has to be improved. Because fallback-jobs, we need go through it and run queries with or without their corresponding fallback.

lib/service/database.js
rochoa: What does happen if I send all the queries as objects? Like:

```
var queries = [
  {query: 'select ...'},
  {query: 'select ...', onerror: 'update...'},
  {query: 'select ...',}
]
```

So we don't need `lib/service/job.js`. I think that would also help to keep the inline execution simpler.

dgaubert:
🙏 🙏 🙏   👀!

lib/service/database.js
```diff
@@ -280,20 +280,35 @@ DatabaseService.prototype.queueAnalysisOperations = function(analysis, callback)
...
+                    query: updateNodeAtAnalysisCatalogQuery(nodeToUpdate.id(), Node.STATUS.RUNNING),
+                    onerror: updateNodeAtAnalysisCatalogQuery(nodeToUpdate.id(), Node.STATUS.FAILED)
+                });
+                asyncQueries.push({
+                    query: populateCacheTableQuery(targetTableName, nodeToUpdate.sql()),
+                    onerror: updateNodeAtAnalysisCatalogQuery(nodeToUpdate.id(), Node.STATUS.FAILED)
+                });
             }
-            asyncQueries.push(updateNodeAtAnalysisCatalogQuery(nodeToUpdate.id(), Node.STATUS.READY));
+            asyncQueries.push({
```
rochoa: What about: 

```diff
diff --git a/lib/service/database.js b/lib/service/database.js
index bccbd12..35c29c4 100644
--- a/lib/service/database.js
+++ b/lib/service/database.js
@@ -291,21 +291,20 @@ DatabaseService.prototype.queueAnalysisOperations = function(analysis, callback)
                 var targetTableName = nodeToUpdate.getTargetTable();
                 asyncQueries.push({
                     query: deleteFromCacheTableQuery(targetTableName),
+                    onsuccess: updateNodeAtAnalysisCatalogQuery(nodeToUpdate.id(), Node.STATUS.RUNNING),
                     onerror: updateNodeAtAnalysisCatalogQuery(nodeToUpdate.id(), Node.STATUS.FAILED)
                 });
                 asyncQueries.push({
-                    query: updateNodeAtAnalysisCatalogQuery(nodeToUpdate.id(), Node.STATUS.RUNNING),
+                    query: populateCacheTableQuery(targetTableName, nodeToUpdate.sql()),
+                    onsuccess: updateNodeAtAnalysisCatalogQuery(nodeToUpdate.id(), Node.STATUS.READY),
                     onerror: updateNodeAtAnalysisCatalogQuery(nodeToUpdate.id(), Node.STATUS.FAILED)
                 });
+            } else {
                 asyncQueries.push({
-                    query: populateCacheTableQuery(targetTableName, nodeToUpdate.sql()),
+                    query: updateNodeAtAnalysisCatalogQuery(nodeToUpdate.id(), Node.STATUS.READY),
                     onerror: updateNodeAtAnalysisCatalogQuery(nodeToUpdate.id(), Node.STATUS.FAILED)
                 });
             }
-            asyncQueries.push({
-                query: updateNodeAtAnalysisCatalogQuery(nodeToUpdate.id(), Node.STATUS.READY),
-                onerror: updateNodeAtAnalysisCatalogQuery(nodeToUpdate.id(), Node.STATUS.FAILED)
-            });
         });

         if (asyncQueries.length) {
```

dgaubert: Nice!

PR #35, opened 2016-06-01
adding dot density as a camshaft analysis

stuartlynn:
Adding Dot Density analysis definition (needs crankshaft to be installed) 

Can either take as params _points_ in which case it just generates that number of points within a polygon 

or _points_col_name_ which uses a column on the input query to populate the geometry and optionally _category_col_name_ which assigns those points a category in the returned table

@rochoa How does this look. Does the kind of variable overloading here that I am doing look good? Trying to get a feel of how we handle inputs to more complicated queries.

Also added two examples to the viewer

### DIFF
--- a/lib/node/nodes/dot-density.js
+++ b/lib/node/nodes/dot-density.js
@@ -36,7 +36,7 @@ var queryTemplate = dot.template([
     'SELECT cdb_crankshaft.cdb_dot_density(_polygons.the_geom, {{=it.points_exp}}::INT) as the_geom',
     '{{? it.category_col_name}} , {{=it.category_col_name}}  as category{{?}}',
     'FROM _polygons'].join('\n')
-)
+);
 
 function query(it) {
     it.points_exp = it.points_col_name ? it.points_col_name : it.points;
--- a/lib/node/nodes/dot-density.js
+++ b/lib/node/nodes/dot-density.js
@@ -1,0 +1,44 @@
+'use strict';
+
+var dot = require('dot');
+dot.templateSettings.strip = false;
+
+var Node = require('../node');
+
+var TYPE = 'dot_density';
+var PARAMS = {
+    source: Node.PARAM.NODE(Node.GEOMETRY.POLYGON),
+    points_col_name: Node.PARAM.NULLABLE(Node.PARAM.STRING),
+    category_col_name: Node.PARAM.NULLABLE(Node.PARAM.STRING),
+    points: Node.PARAM.NULLABLE(Node.PARAM.NUMBER)
+};
+
+var DotDensity = Node.create(TYPE, PARAMS);
+
+module.exports = DotDensity; 
+module.exports.TYPE = TYPE;
+module.exports.PARAMS = PARAMS;
+
+DotDensity.prototype.sql = function() {
+    return query({
+        query: this.source.getQuery(),
+        points: this.points,
+	points_col_name: this.points_col_name,
+	category_col_name: this.category_col_name
+    });
+};
+
+var queryTemplate = dot.template([
+    'WITH',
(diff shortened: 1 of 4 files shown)


lib/node/nodes/dot-density.js
```diff
@@ -0,0 +1,44 @@
...
+
+var dot = require('dot');
+dot.templateSettings.strip = false;
+
+var Node = require('../node');
+
+var TYPE = 'dot_density';
+var PARAMS = {
+    source: Node.PARAM.NODE(Node.GEOMETRY.POLYGON),
+    points_col_name: Node.PARAM.NULLABLE(Node.PARAM.STRING),
```
rochoa: Shouldn't all these params be mandatory? If they are not provided the analysis won't work, will it?

stuartlynn: Yeah I wanted to discuss this with you to try and figure out what the best practice is. So Dot Density can run in a number of ways. It always needs to have a geometry column but then you can either 

- Pass just a plane number of points to populate each geometry by : the points param. Example [here](https://github.com/CartoDB/camshaft/blob/b639b5e8c22cd2d999f3e38bea86cf6d80de1647/examples/viewer/examples.js#L529) 
- Pass the name of a points column, in which case Dot Density will populate each geometry with the corresponding number of points in that column. Example [here](https://github.com/CartoDB/camshaft/blob/b639b5e8c22cd2d999f3e38bea86cf6d80de1647/examples/viewer/examples.js#L550) 
- Pass the name of a points column + the name of a category column so that all points that get generated for that geometry get an additional column called category which corresponds to the category column name. Same example as before 

So in a sense we have  ```dot_density(geometry, points)```, ```dot_density(geometry, points_col_name)``` or ```dot_density(geometry,points_col_name,category_col_name)```  So we need at least one of points or points_col_name. 

The alternative would be having something like a Node.PARAM.VALUE_OR_COLUMN_NAME parameter that could accept either a value or a column name.

What do you think?

rochoa: I would like to add @viddo to this conversation as this might affect a lot they way he is building the editor components to handle these optional params.

Other alternative is going more granular and have different analyses for different options:

1. For points
1. Points column + optional category column.

stuartlynn: Thats definitely one way to go but I think that  having two dot density options in the editor would be confusing to a user. This feels more to me like a single analysis with multiple different options for input.

viddo: @rochoa this sounds similar to the [buffer analysis and its not-always-optional params](https://github.com/CartoDB/cartodb/issues/7502#issuecomment-223015673), right? If so I think this isn't something that could/should be tackled separately as it's not unique for this dot density analysis. 

For the moment these kind of restrictions are implemented by a developer on a per-analysis/use-case basis anyway, so the only thing we need is to know about these restrictions/combinations when implementing the form/validation rules. But if we could define them in the reference that would 👌 ofc, but like wrote above, let's take that conversation elsewhere.

rochoa: @stuartlynn have in mind that in some cases the editor will take care of that and will display just one analysis although it might be using two behind the scenes. Think about _area of influence_ where we use `trade-area` or `buffer` analyses depending whether is time or distance.

IMO we should split this analysis into two as I proposed.

stuartlynn:
Hey guys. Just checking where we ended up with this? It sounds like we can merge it in and design the interface to deal with the issues here?

rochoa:
This kind of analyses, general purpose or ad-hoc ones, should use the [`deprecated-sql-function`](https://github.com/CartoDB/camshaft/blob/0.54.0/docs/deprecated-sql-function.md).

PR #36, opened 2016-06-03
Adding a function to apply an arbitrary SQL statement to the previous s…

stuartlynn:
…tate

@ohasselblad @rochoa Adding this for discussion. Basically allows you to specify an arbitrary SQL statement to the source table.

### DIFF
--- a/examples/viewer/examples.js
+++ b/examples/viewer/examples.js
@@ -44,6 +44,14 @@ var sourceLaLatina = {
     }
 };
 
+var SQLDef = {
+    type: 'SQL',
+    params: {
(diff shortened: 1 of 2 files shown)


ohasselblad:
This is a big thumbs up from me for the same reasons that people love CartoDB for exposing SQL, we can expose user defined analyses so people can put in their custom analysis along with the ones we are defining. This could also be a good way to see what gaps are to be filled in the analyses that we're building.

andrewxhill:
where is the /nodes/sql file though?

rochoa:
This is superseded by the [`deprecated-sql-function`](https://github.com/CartoDB/camshaft/blob/0.54.0/docs/deprecated-sql-function.md).

PR #43, opened 2016-06-06
Merge analysis

dgaubert:
first approach

cc @rochoa


lib/node/nodes/merge.js
```diff
@@ -0,0 +1,68 @@
...
+    '{{=it.input_right_alias}} AS (',
+    ' {{=it.input_right}}',
+    ')',
+    'SELECT ' +
+    '  {{=it.columns}},',
+    '  ST_Union({{=it.input_left_alias}}.the_geom, {{=it.input_right_alias}}.the_geom) AS the_geom',
+    'FROM {{=it.input_left_alias}} {{?it.left_join}}LEFT{{?}} JOIN {{=it.input_right_alias}}',
+    'ON {{=it.input_left_column_on}} = {{=it.input_right_column_on}}',
+    'AND ST_Intersects({{=it.input_left_alias}}.the_geom, {{=it.input_right_alias}}.the_geom)'
+].join('\n'));
```
dgaubert: From mockups I understood that analysis has to join by _no-geom-column_ and perform the union of their geometries. My question is, those geometries have to intersect between them?

rochoa: I'm not sure if we want to union the geometries. @saleiva what should we do with geometries: keep both or just use the one from the left?

lib/node/nodes/join.js
```diff
@@ -0,0 +1,65 @@
+'use strict';
+
+var dot = require('dot');
+dot.templateSettings.strip = false;
+
+var Node = require('../node');
+
+var INPUT_LEFT_ALIAS = '_cdb_analysis_input_left';
```
dgaubert: I'm not totally happy with the naming used in this query, specially with aliases. @rochoa, what do you think?

Issue #38, opened 2016-06-03
Perfect Sales Territory - K means with Weights Analysis 


stuartlynn:
Create a Camshaft/Crankshaft analysis that takes a list of points and weights and calculates n weighted cluster centers for those points. 

Use sk-learn with modified distance formula so that we can find the weighted centroid rather than the regular one, something like: 

$d = ((x-x)^2 + (y-y)^2 )/w$

rochoa:
What would you expect from this analysis as output? A new polygon geometry or a column with categories?

cc @javisantana

javisantana:
confirmed with @andrewxhill, a new column with the cluster number (or something like that)

andrewxhill:
Inputs:
* Many points in a table
* N, input of desired clusters

Output
* N point geometries (centers of clusters)
* N cluster IDs

javisantana:
what is the ETA?

stuartlynn:
Should have a crankshaft and camshaft PR ready today

javisantana:
ok, nice

andrewxhill:
Sorry.. talking with stuart my answer is actually wrong for the kmeans bit

Inputs:

Many points in a table
N, input of desired clusters

Output

Many points table
Column of 1-N cluster IDs

So a table with 100 points will return a table with 100 points plus a cluster id column

stuartlynn:
We also have another helper function that will take those clusters and produce the weighted centers of those N clusters based on some other column 

Video of the process is here :
https://files.slack.com/files-pri/T024FUQQ8-F1ES9V6BT/download/store_locator_analysis.mov

stuartlynn:
This should be in place on both the crankshaft and camshaft sides

PR #44, opened 2016-06-08
adding kmeans and weighted centroid nodes

stuartlynn:
Adds the nodes for #38 kmeans and weighted centroid

### DIFF
--- a/examples/viewer/examples.js
+++ b/examples/viewer/examples.js
@@ -69,8 +69,8 @@ var WeightedCentroidDefinition={
     type: 'weighted_centroid',
     params:{
         source: customersSourceDef,
-        weight_col: 'customer_v',
-        category_col:'cat'
+        weight_column: 'customer_v',
+        category_column:'cat'
     }
 }
 
--- a/examples/viewer/examples.js
+++ b/examples/viewer/examples.js
@@ -26,7 +26,7 @@ var sourceExpensiveRentListings = {
     params: {
         query: 'select * from airbnb_madrid_oct_2015_listings where price > 100'
     }
-}
+};
 
 var sourceBarrios = {
     id: 'barrios-source',
@@ -40,7 +40,7 @@ var sourceLaLatina = {
     id: 'barrios-source',
     type: 'source',
     params: {
-        query: "select * from barrios where codbarrio like '10%'"
+        query: 'select * from barrios where codbarrio like \'10%\''
     }
 };
 
@@ -48,7 +48,7 @@ var customersSourceDef = {
     id: 'customersSource',
     type: 'source',
     params: {
-        query: "select *, category::integer cat  from customers_3"
+        query: 'select *, category::integer cat  from customers_3'
     }
 };
 
@@ -66,20 +66,20 @@ var tradeAreaDefinition = {
 
 var WeightedCentroidDefinition={
     id: 'weightedCentroid',
-    type: 'weighted_centroid',
+    type: 'weighted-centroid',
     params:{
         source: customersSourceDef,
         weight_column: 'customer_v',
         category_column:'cat'
     }
-}
+};
 
 var KMeansDefinition ={
     id: 'kmeans',
     type: 'kmeans',
     params:{
       source: customersSourceDef,
-      no_clusters : 6
+      clusters : 6
     }
 };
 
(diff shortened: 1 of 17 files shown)


rochoa:
Let me help with the travis thingy.

rochoa:
Wouldn't be better to have different analyses in different PR?

stuartlynn:
Totally. I can separate them out. These are both for the usecase we are trying to rush together for the smoke tests so I just lumped them together for now. Happy to break them out in future

lib/node/nodes/kmeans.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+var Node = require('../node')
+var dot = require('dot')
+
+var TYPE = 'kmeans';
+
+var PARAMS = {
+    source : Node.PARAM.NODE(Node.GEOMETRY.POINT),
+    no_clusters : Node.PARAM.NUMBER
```
rochoa: What about using just `clusters`?

examples/viewer/examples.js
```diff
@@ -56,6 +64,25 @@ var tradeAreaDefinition = {
     }
 };
 
+var WeightedCentroidDefinition={
+    id: 'weightedCentroid',
+    type: 'weighted_centroid',
+    params:{
+        source: customersSourceDef,
+        weight_col: 'customer_v',
```
rochoa: I would use `_column` to be consistent with already existing analyses.

stuartlynn: Sounds good

rochoa:
Don't worry, next time we will split them. Keep it like this!

rochoa:
I pushed some fixes so travis stops complaining 😈 .

stuartlynn:
Thanks!
On Wed, Jun 8, 2016 at 1:02 PM Raul Ochoa <notifications@github.com> wrote:

> I pushed some fixes so travis stops complaining 😈 .
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/CartoDB/camshaft/pull/44#issuecomment-224658462>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAMzrr5TbItaBDnH4gqgwMKfZtEBVztHks5qJvWigaJpZM4IxEjJ>
> .
>
-- 
*Stuart Lynn*
Map Scientist

*CartoDB—Map Your World's Data*
143 Roebling St 2A, New York, 11211, USA
T: +3123940797  Twitter: @stuart_lynn
W: www.cartodb.com

Live sessions to learn CartoDB

lib/node/nodes/kmeans.js
```diff
@@ -0,0 +1,35 @@
...
+var dot = require('dot');
+
+var TYPE = 'kmeans';
+
+var PARAMS = {
+    source : Node.PARAM.NODE(Node.GEOMETRY.POINT),
+    clusters : Node.PARAM.NUMBER
+};
+
+var KMeans  = Node.create(TYPE, PARAMS, {cache: false});
```
rochoa: We will need to `{cache: true}` this analysis as it need to be run at whole dataset level, otherwise it will run at per [meta]tile level so it won't make sense at all.

`cache: true`
screen shot 2016-06-09 at 13 30 17

`cache: false`
screen shot 2016-06-09 at 13 30 54


cc @javisantana

lib/node/nodes/weighted-centroid.js
```diff
@@ -0,0 +1,36 @@
...
+
+var TYPE = 'weighted-centroid';
+
+var PARAMS = {
+    source : Node.PARAM.NODE(Node.GEOMETRY.POINT),
+    weight_column : Node.PARAM.STRING,
+    category_column : Node.PARAM.STRING
+};
+
+var WeightedCentroid  = Node.create(TYPE, PARAMS, {cache: false});
```
rochoa: I guess we will have the very same problem with this one, right?

lib/node/nodes/weighted-centroid.js
```diff
@@ -0,0 +1,36 @@
...
+
+module.exports = WeightedCentroid;
+module.exports.TYPE = TYPE;
+module.exports.PARAMS = PARAMS;
+
+var weightedCentroidTemplate = dot.template([
+    'SELECT the_geom , class',
```
rochoa: Should we select the rest of the columns, right?

lib/node/nodes/weighted-centroid.js
stuartlynn: Happy to mark this and the kmeans as cache = true

lib/node/nodes/weighted-centroid.js
stuartlynn: This is more of an aggregation , it operates on a table with say 100 points and 3 classes and produces 3 points. We could add other aggregate columns, like the number of points that centroid is generated from or for example some aggregate over a column of all the points that belong to that weighted mean but I dont think its appropriate to bring back columns from the original table

lib/node/nodes/kmeans.js
stuartlynn: Yup sounds great.I was going to ask how we avoid the meta tile issue

javisantana: This is really important, all the analysis that need knoledge about the full dataset should:

- set cached: true
- rethinking about them because we should try to avoid them (they are not easily parallelizable)

stuartlynn: Just pushed a commit setting the caches to True. What should be the strategy for analysis like this that need info from the entire table? I totally agree with your concerns but there are a lot of really powerful analysis methods that are going to need full knowledge of the table, would be good to have a brainstorm at some point about how we might be able to write them in a way thats more efficient and can be made parallelizable

lib/node/nodes/weighted-centroid.js
rochoa: Oh, thanks. Understood.

What about producing a cartodb_id column? Other analyses might need that. For instance, and I know it's a silly example, kmeans require to have a cartodb_id column: if I apply a `kmeans` over the result of `weighted-centroid` it will fail because `weighted-centroid` is not producing a cartodb_id column.

stuartlynn: That sounds good. Just a new sequential id ?

rochoa: I'm not fully sure, but I just realised that this might happen in other analyses as well. Actually the buffer/trade-area analyses don't return cartodb_id depending on some options.

So feel free to skip that for now. I will take care of the cartodb_id at a higher level in camshaft.

lib/node/nodes/kmeans.js
stuartlynn: Also this analysis is just using lat lng distances just now so we probably want to keep examples to city level otherwise its going to become less accurate

javisantana: for the moment we just need to set `cache: true` so we avoid postgres planner to make wrong decisions.

In the future we will see

rochoa:
:ship:ing 🇮🇹 . 🎉  thanks everyone involved!!

PR #62, opened 2016-06-16
[work in progress] distance weighted knn prediction

ohasselblad:
As requested by @javisantana, I'm handing this PR over to @rochoa.  

The function needs to be swapped out with `CDB_SpatialInterpolation` which has a [different signature](https://github.com/CartoDB/crankshaft/blob/faa899cf8707adab4fc49834afa0a4d566abb481/src/pg/sql/08_interpolation.sql#L27-L35). This function is a more general version of the one I wrote. This PR depends on https://github.com/CartoDB/crankshaft/pull/71 for the deploy of the new function.

cc @AbelVM please let us know if there's something missing here.

Inputs need to be:
1. source geoms (array of geometries or table/analysis step where geometries come from)
2. source values (array of numeric values at each of the geoms in the first step)
3. point column (values for predicting a new measurement from)
4. method for comparison (one of: nearest neighbor, barycentric, or inverse-distance weighting) -- optional, default to nearest inverse-distance weighting (option 2),
5. if inverse-distance weighting, pick number of neighbors 
6. if inverse-distance weighting, pick decay strength (0 gives average of k nearest neighbors, 1 is weak decay, 2 is stronger decay, etc. etc.)

Output:
1. Interpolated value (numeric)

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -14,7 +14,8 @@ var nodes = {
     PopulationInArea: require('./nodes/population-in-area'),
     Source: require('./nodes/source'),
     TradeArea: require('./nodes/trade-area'),
-    WeightedCentroid: require('./nodes/weighted-centroid')
+    WeightedCentroid: require('./nodes/weighted-centroid'),
+    weightedKnn: require('./nodes/distance-weighted-knn')
 };
 
 module.exports = nodes;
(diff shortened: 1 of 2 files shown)


stuartlynn:
Hey @rochoa we just did a code review on this so whenever https://github.com/CartoDB/crankshaft/pull/63 gets merged this should be good to go as well.

ohasselblad:
Closing. Will replace with a new PR eventually.

PR #90, opened 2016-06-24
Analysis routing point to point

dgaubert:

- [x] Outputs a `cartodb_id numeric` column. (_if source node has that column_)
- [x] Avoids using CTEs for join operations when result is not cached. (_actually is using them, but is cached_)

cc @rochoa


lib/node/nodes/routing-to-single-point.js
```diff
@@ -0,0 +1,33 @@
...
+
+var routingToSinglePointQueryTemplate = Node.getSqlTemplateFn('routing-to-single-point');
+
+var TYPE = 'routing-to-single-point';
+var PARAMS = {
+    source: Node.PARAM.NODE(Node.GEOMETRY.POINT),
+    mode: Node.PARAM.ENUM('car', 'walk', 'bicycle', 'public_transport'),
+    destination_longitude: Node.PARAM.NUMBER(),
+    destination_latitude: Node.PARAM.NUMBER(),
+    units: Node.PARAM.ENUM('kilometers', 'miles'),
```
rochoa: I would make this to default to kilometres.

PR #96, opened 2016-06-28
WIP - Stubbing out segmentation node

stuartlynn:
This is a placeholder for the segmentation analysis that works with this crankshaft pull request: 
https://github.com/CartoDB/crankshaft/pull/70

The function you want to wrap up is this one : 

https://github.com/CartoDB/crankshaft/blob/5f98735ce97bb040d1233252e2fce40a175de728/doc/12_segmentation.md#cdb_createandpredictsegmenttarget-numeric-train_features-numeric-prediction_features-numeric-prediction_ids-numeric

To use it you need to use the pyagg helper function to get the data in to the segmentation. See the example here: 

https://github.com/CartoDB/crankshaft/blob/5f98735ce97bb040d1233252e2fce40a175de728/doc/12_segmentation.md#example-usage-1

Open questions are how much of the GradientBoosting hyper-parameters do we expose in the editor. 

### DIFF
--- a/lib/node/nodes/segmentation.js
+++ b/lib/node/nodes/segmentation.js
@@ -1,0 +1,19 @@
+'use strict';
+
+var Node = require('../node');
+
+var TYPE = 'market-prediction';
+var PARAMS = {
+    source: Node.PARAM.NODE(Node.GEOMETRY.POLYGON),
+    target: Node.PARAM.NODE(Node.GEOMETRY.POLYGON)
+   };
+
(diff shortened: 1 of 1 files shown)


andrewxhill:
cc @rochoa @javisantana

rochoa:
This kind of analyses, general purpose or ad-hoc ones, should use the [`deprecated-sql-function`](https://github.com/CartoDB/camshaft/blob/0.54.0/docs/deprecated-sql-function.md).

PR #112, opened 2016-07-03
Markov

javisantana:
adds markov analysis

I also added a check for auto node loading

cc @rochoa


saleiva:
Did the form work?

javisantana:
@saleiva yes, I think we can remove the "output col" since it's many columns and with fixed names (as moran does)

PR #116, opened 2016-07-04
make kmeans analysis consistent between runs

javisantana:
cc @rochoa what do you think?


rochoa:
Let's ship it like this. We will have to evaluate how to generalise the solution.

PR #120, opened 2016-07-05
Now SQL functions are fully qualified

### LINKED ISSUE #113: SQL functions must be fully qualified
Using the correct schema from the extension they belong to.

dgaubert:
Fixes #113

### DIFF
--- a/test/setup.js
+++ b/test/setup.js
@@ -14,6 +14,7 @@ var DATABASE_NAME = testConfig.db.dbname;
 
 before(function setupTestDatabase(done) {
     var fixturePaths = [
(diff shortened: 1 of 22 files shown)


test/setup.js
```diff
@@ -37,6 +37,10 @@ before(function setupTestDatabase(done) {
             function createPostgisExtension(stdout, stderr, callback) {
                 exec('psql -d ' + DATABASE_NAME + ' -c "CREATE EXTENSION postgis;"', callback);
             },
+            function createDataServicesClientSchema(stdout, stderr, callback) {
```
rochoa: I would put this in a fixture file.

dgaubert: Or as first line in every data-services' fixture file.

PR #121, opened 2016-07-05
Add density column

dgaubert:


lib/node/nodes/aggregate-intersection.js
```diff
@@ -41,10 +51,20 @@ function setAliasPrefixToColumnNames(columnNames, alias) {
     }).join(', ');
 }
 
+var densityQueryTemplate = Node.template([
+    ',',
+    '{{=it.aggregateFunction}}({{=it.qualifiedAggregateColumn}})',
+    '/',
+    // use 2.6e-06 as minimum area (tile size at zoom level 31)
+    'GREATEST(0.0000026, ST_Area((ST_Transform(_cdb_analysis_source.the_geom, 4326))::geography))',
```
rochoa: Why do you transform to 4326? `the_geom` must be 4326 already.

dgaubert: @javisantana told me, because `2.6e-06` is the minimum area (zoom 31) for the pixel size (if I understood correctly)

lib/node/nodes/aggregate-intersection.js
```diff
@@ -41,10 +51,20 @@ function setAliasPrefixToColumnNames(columnNames, alias) {
...
+    'as aggregate_density',
+].join(' '));
+
 var queryAggregateTemplate = Node.template([
     'SELECT',
     '  {{=it.sourceColumns}},',
     '  {{=it.aggregateFunction}}({{=it.qualifiedAggregateColumn}}) as {{=it.aggregateFinalColumnName}}',
+    '  {{=it.aggregateDensity}}',
```
dgaubert: done!

PR #124, opened 2016-07-05
Add geocoder provider in geocode street analysis

iriberri:
Not adding a new analysis but editing an existent one, so I'm ignoring the checklist below (my additions follow the naming conventions).

The node now accepts the geocoder provider as a parameter. If mapzen want to be used then `provider` should be `mapzen`. Otherwise it will use the default geocoder provider for the user.

====

### DIFF
--- a/lib/node/nodes/georeference-city.js
+++ b/lib/node/nodes/georeference-city.js
@@ -7,7 +7,8 @@ var PARAMS = {
     source: Node.PARAM.NODE(Node.GEOMETRY.ANY),
     city: Node.PARAM.STRING(),
     admin_region: Node.PARAM.STRING(),
-    country: Node.PARAM.STRING()
+    country: Node.PARAM.STRING(),
+    provider: Node.PARAM.STRING()
 };
 
 var GeoreferenceCity = Node.create(TYPE, PARAMS, { cache: true });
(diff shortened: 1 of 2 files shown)


lib/node/nodes/georeference-city.js
```diff
@@ -7,7 +7,8 @@ var PARAMS = {
     source: Node.PARAM.NODE(Node.GEOMETRY.ANY),
     city: Node.PARAM.STRING(),
     admin_region: Node.PARAM.STRING(),
-    country: Node.PARAM.STRING()
+    country: Node.PARAM.STRING(),
+    provider: Node.PARAM.STRING()
```
javisantana: use NULLABLE here, like

```
Node.PARAM.NULLABLE(Node.PARAM.STRING(), 'mapzen')
```

iriberri:
Just applied the feedback @javisantana, thank you! Tests are passing locally, let's see what Travis says :-)

javisantana:
cc @dgaubert

lib/node/nodes/georeference-city.js
```diff
@@ -7,7 +7,8 @@ var PARAMS = {
     source: Node.PARAM.NODE(Node.GEOMETRY.ANY),
     city: Node.PARAM.STRING(),
     admin_region: Node.PARAM.STRING(),
-    country: Node.PARAM.STRING()
+    country: Node.PARAM.STRING(),
+    provider: Node.PARAM.NULLABLE(Node.PARAM.STRING(), 'mapzen')
```
dgaubert: Who is going to provide this param?. Should the user choose what 3rd party service we use internally?

javisantana: yes

dgaubert:
LGTM

👍

lib/node/nodes/georeference-city.js
rochoa: We should make this an ENUM so we constraint it to the valid params.

iriberri: Makes sense. Will iterate over that and name the parameter to use the generic function (no provider defined in the signature, uses whatever the user has configured in his/her account) as `user_default` or similar. Thanks!

rochoa: This is not critical at all :-). We can live with it as it is right now.

test/fixtures/cdb_geocoder.sql
```diff
@@ -101,6 +101,18 @@ RETURNS Geometry AS $$
   END
 $$ LANGUAGE plpgsql;
 
+CREATE OR REPLACE FUNCTION cdb_mapzen_geocode_street_point(searchtext TEXT, city TEXT DEFAULT NULL, state_province TEXT DEFAULT NULL, country TEXT DEFAULT NULL)
```
dgaubert: I've just realised that `cdb_mapzen_geocode_street_point` fixture is not used by `GeoreferenceCity` analysis, that uses `cdb_mapzen_geocode_namedplace_point`. The same for other providers. @iriberri, which one is the good one?

iriberri: Oh ****.  It's geocode_street. Pushing a fix right now.

lib/node/nodes/georeference-city.js
iriberri: Added in https://github.com/CartoDB/camshaft/pull/125 as I totally messed up yesterday with the functions. Oops.

PR #126, opened 2016-07-07
Mapzen isocrones in analysis

iriberri:
~~Do not merge yet. I'm in the process of releasing the functions included here, it won't take much but it's not ready yet.~~

Similarly to the street geocoder analysis, I'm adding providers in the trade areas one. In this scenario, we have only two providers: `heremaps` and `mapzen`. I'm setting `mapzen` as the default. Please, take into account that `mapzen` function is slower than the `heremaps` service.

### DIFF
--- a/lib/node/nodes/sql/trade-area-dissolved.sql
+++ b/lib/node/nodes/sql/trade-area-dissolved.sql
@@ -4,7 +4,7 @@ _cdb_analysis_source_points AS (
 ),
 _cdb_analysis_isochrones AS (
   SELECT
-    cdb_dataservices_client.cdb_isochrone(
+    cdb_dataservices_client.{{=it.provider_function}}(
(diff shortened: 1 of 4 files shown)


iriberri:
cc @dgaubert 

I didn't change the current analysis examples to add the provider parameter. Should I?

iriberri:
Functions are in production now :+1:

iriberri:
cc @javisantana, I forgot ;)

dgaubert:
> I didn't change the current analysis examples to add the provider parameter. Should I?

Not is needed, because the new param has a default value.

BTW I've just bamped trade-area version, to avoid caching existing nodes with same params.

iriberri:
Thanks @dgaubert!

dgaubert:
Let's wait for green... and merge!

PR #133, opened 2016-07-11
Added node_id param to errors in node creation

dgaubert:
To provide more context to final user, this PR just add the node_id in the error object to be used by camshaft's client.

### DIFF
--- a/lib/node/node.js
+++ b/lib/node/node.js
@@ -337,19 +337,15 @@ module.exports.create = function createNode(type, expectedParams, options) {
             var validator = expectedParams[expectedParamName];
 
             if (!validator.nullable && !this.params.hasOwnProperty(expectedParamName)) {
-                var requiredParamError = new Error('Missing required param "' + expectedParamName + '"');
-                requiredParamError.node_id = this.id();
-                throw requiredParamError;
+                throw new Error('Missing required param "' + expectedParamName + '"');
             }
 
             if (!validator.isValid(param)) {
-                var isNotValidError = new Error(
+                throw new Error(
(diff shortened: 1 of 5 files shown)


lib/node/node.js
```diff
@@ -337,15 +337,19 @@ module.exports.create = function createNode(type, expectedParams, options) {
...
             }
 
             if (!validator.isValid(param)) {
-                throw new Error(
+                var isNotValidError = new Error(
                     'Invalid type for param "' + expectedParamName + '",' +
                         ' expects "' + validator.expects + '" type,' +
                         ' got `' + JSON.stringify(param) + '`'
                 );
+                isNotValidError.node_id = this.id();
```
rochoa: This is the internal ID of the node; it's not the id the user provided. Exposing this id doesn't help the end user because the id the user provided is not this one.

We need to rethink the naming; probably we need to rename the internal node id to `node.ref` or something similar. Or the external to `node.name`.

dgaubert:
Updated! please @rochoa, CR.

lib/workflow/factory.js
```diff
@@ -82,6 +82,7 @@ Factory.prototype._create = function (definition, depth, callback) {
         try {
             node = new NodeClass(self.user, definition.params);
         } catch (err) {
+            err.node_id = definition.id;
```
rochoa: This will cover most of the cases when input params are invalid, but this won't solve other cases when something fails later in the factory, for instance when creating a cache table.

Let's ship it and we will work on improving other cases :-).

dgaubert: 👌

PR #140, opened 2016-07-14
fix postal code params

### LINKED ISSUE #138: Analysis of geocoding analyses
Hey there! I'm taking a look at the current geocoder analyses from the Data Services team point of view. I hope this helps!

## Administrative regions 
Analysis [here](https://github.com/CartoDB/camshaft/blob/master/lib/node/nodes/georeference-admin-region.js).

For this category, DS API has a function per each administrative level. Right now we have 2 of them: [admin0](https://carto.com/docs/carto-engine/dataservices-api/geocoding-functions/#cdbgeocodeadmin0polygoncountryname-text) (countries) and [admin1](https://carto.com/docs/carto-engine/dataservices-api/geocoding-functions/#cdbgeocodeadmin1polygonadmin1name-text) (states, provinces)

### Country geocoding

* Parameters:
  * **country** (mandatory): from a column value 

##### DS API function signatures:
* `cdb_geocode_admin0_polygon(country_name text)`

The result is always a polygon.

### Admin. level 1 (provinces/states) geocoding

* Parameters:
  * **admin1_name** (mandatory): from a column value
  * **country_name** (optional): from a column value or free text

##### DS API function signatures:
* `cdb_geocode_admin1_polygon(admin1_name text)`
  * To be used when we only have admin1_name, sending a `null` as country will not bring back the desired result
* `cdb_geocode_admin1_polygon(admin1_name text, country_name text)`
  * To be used when we have the country (in a column or as free text)

The result of these functions is always a polygon.

### Spotted issues
* [x] Currently there's no geocoder for countries. https://github.com/CartoDB/camshaft/issues/128#issuecomment-231720005
* [x] Admin1 geocoder should use different functions from the Data Services API depending on the parameters
* [x] For administrative regions of level 1, the country parameter should be optional. 
* [x] For administrative regions of level 1, the country parameter could be free text. Already reported at https://github.com/CartoDB/camshaft/issues/137

## IP Addresses

* Parameters
  * **IP addr** (mandatory): from a column value

##### DS API function signatures:
* `cdb_geocode_ipaddress_point(ip_address text)`
  * To be used always, there are no special cases for this geocoding

This function always return points.

### Spotted issues
None 👍 😄 

## City names

* Parameters
  * **city name** (mandatory): from a column value
  * **admin1 name** (optional): from a column value or free text
  * **country name** (optional): from a column value or free text

##### DS API function signatures:
* `cdb_geocode_namedplace_point(city_name text)`
  * To be used when we just have the city name
* `cdb_geocode_namedplace_point(city_name text, country_name text)`
  * To be used when we have city + country
* `cdb_geocode_namedplace_point(city_name text, admin1_name text, country_name text)`
  * To be used when we have city + region + country

### Spotted issues
* [x] The function with three parameters is always being used and it will not get a response with null values as parameters
* [x] Similarly as with other analysis, region and country parameters can be optional
* [x] Similarly as with other analysis, region and country parameters can be free text

## Street addresses

* Parameters
  * **address** (mandatory): from a column value
  * **city** (optional): from a column value or free text
  * **state/province/region** (optional): from a column value or free text
  * **country** (optional): from a column value or free text

##### DS API function signatures:

* `cdb_geocode_street_point(search_text text, [city text], [state text], [country text])`
  * To be used always. This function works in a different way that the rest with respect to the parameters. Null values passed as parameters won't have any bad effect here.

These functions always return points.

### Spotted issues
None 👍 😄 

But a Note:
Mapzen geocoder works in a special way and works much better if we send the country code in ISO3. We do this internally through the API, but for this we need the `country` parameter to be filled -- [we convert it to ISO3](https://github.com/CartoDB/dataservices-api/blob/261fff787373974b481a6a5716b6c841126ffa5d/server/extension/sql/20_geocode_street.sql#L149). The current implementation of this analysis is respecting this, but it's something to take into account if something is to be changed. :)

## Postal codes

* Parameters
  * **postal code** (mandatory): from a column value
  * **country** (mandatory): from a column value or free text

##### DS API function signatures:
* `cdb_geocode_postalcode_polygon(postal_code text, country_name text)`
  * To be used when we want polygon results. This function only has coverage on USA, FRA, CAN and AUS.
* `cdb_geocode_postalcode_point(postal_code text, country_name text)`
  * To be used when we want point results. This function has coverage in most part of the world.

### Spotted issues
* [x] Only the `polygon` function is being used at this moment in the analysis. There should be a new parameter with the type of geometry (point or polygon) which should determine which of the two functions above is called. Parameters are the same for both but the result varies in terms of the geometry.

dgaubert:
Fixes #138 partially:
  - Adds param to choose geometry type output (point or polygon).
  - Adds param to indicate if param is a column value or custom text.

### DIFF
--- a/lib/node/nodes/georeference-postal-code.js
+++ b/lib/node/nodes/georeference-postal-code.js
@@ -5,7 +5,7 @@ var Node = require('../node');
 var TYPE = 'georeference-postal-code';
 var PARAMS = {
     source: Node.PARAM.NODE(Node.GEOMETRY.ANY),
-    target_geometry_type: Node.PARAM.NULLABLE(
+    output_geometry_type: Node.PARAM.NULLABLE(
         Node.PARAM.ENUM(Node.GEOMETRY.POINT, Node.GEOMETRY.POLYGON),
         Node.GEOMETRY.POLYGON
     ),
@@ -23,7 +23,7 @@ GeoreferencePostalCode.prototype.sql = function() {
     return queryTemplate({
         source: this.source.getQuery(),
         columns: this.source.getColumns(true).join(', '),
-        geocoder_function: getGecoderFunction(this.target_geometry_type),
+        geocoder_function: getGecoderFunction(this.output_geometry_type),
         postal_code: this.postal_code,
         country: this.country
     });
--- a/lib/node/nodes/georeference-postal-code.js
+++ b/lib/node/nodes/georeference-postal-code.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var Node = require('../node');
+var debug = require('../../util/debug')('analysis:georeference-postal-code');
 
 var TYPE = 'georeference-postal-code';
 var PARAMS = {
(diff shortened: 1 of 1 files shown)


lib/node/nodes/georeference-postal-code.js
```diff
@@ -1,33 +1,74 @@
...
+
+    return 'NULL';
+};
+
+function getGecoderFunction(geometryType) {
+    switch (geometryType){
+        case Node.GEOMETRY.POLYGON:
+            return 'cdb_geocode_postalcode_polygon';
+        case Node.GEOMETRY.POINT:
+            return 'cdb_google_geocode_street_point';
```
iriberri: Ooops, a wild address function appeared! It should be `cdb_geocode_postalcode_point` instead.

PR #146, opened 2016-07-15
Fix point postal code function

iriberri:
I just spotted a typo at: https://github.com/CartoDB/camshaft/pull/140/files#r70935467

Fixing it here :-)

cc @dgaubert


dgaubert:
Ohh! Thanks @iriberri

PR #147, opened 2016-07-18
Use heremaps trade areas

ethervoid:
Due some problems with mapzen trade areas client, we need to change to heremaps for the moment


dgaubert:
👍

rochoa:
Fair enough for now. But we should keep the default value as stable as possible, and change the client using this analysis so it uses the adequate provider value.

PR #151, opened 2016-07-18
Be able to specify source columns in `intersection`

### LINKED ISSUE #148: Be able to specify source columns in `intersection`
It would be great if we could choose those columns we want from polygon dataset.
How difficult would it be @dgaubert?

dgaubert:
Fixes #148

cc @rochoa

### DIFF
--- a/lib/node/nodes/intersection.js
+++ b/lib/node/nodes/intersection.js
@@ -1,27 +1,34 @@
 'use strict';
 
 var Node = require('../node');
-var debug = require('../../util/debug')('analysis:inersection');
+var debug = require('../../util/debug')('analysis:intersection');
 
 var TYPE = 'intersection';
 var PARAMS = {
     source: Node.PARAM.NODE(Node.GEOMETRY.ANY),
-    source_columns: Node.PARAM.NULLABLE(Node.PARAM.ARRAY()),
+    source_columns: Node.PARAM.NULLABLE(Node.PARAM.ARRAY(Node.PARAM.STRING()), []),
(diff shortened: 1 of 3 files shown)


lib/node/nodes/intersection.js
```diff
@@ -1,28 +1,41 @@
...
-    target: Node.PARAM.NODE(Node.GEOMETRY.ANY)
+    source_columns: Node.PARAM.NULLABLE(Node.PARAM.ARRAY(Node.PARAM.STRING()), []),
+    target: Node.PARAM.NODE(Node.GEOMETRY.ANY),
 };
 
-var Intersection = Node.create(TYPE, PARAMS, { cache: true, version: 3 });
+var Intersection = Node.create(TYPE, PARAMS, {
+    cache: true,
+    version: 4,
+    beforeCreate: function (node) {
```
rochoa: Why do you need to retrieve the source_colums here?

dgaubert: To keep previous behaviour, all columns from source. Is it not what we want?

rochoa: But you can do that outside of `beforeCreate`. The `beforeCreate` is supposed to be used when you cannot do something after the node has been created: ignore param for id generation, throw an exception due a specific validation, and similar things.

dgaubert: I didn't know that, thanks for clarifying. Fixing ASAP.

rochoa: Not a high priority thing, just wanted to be sure it was not actually needed. Also wanted to let you know the intentional behaviour for `beforeCreate` 😉 .

PR #153, opened 2016-07-19
Add enrich

AbelVM:

cc @javisantana

### DIFF
--- a/lib/node/nodes/interpolation.js
+++ b/lib/node/nodes/interpolation.js
@@ -1,0 +1,38 @@
+'use strict';
+
+var Node = require('../node');
+
+var TYPE = 'enrich';
+
+var PARAMS = {
+    source: Node.PARAM.NODE(Node.GEOMETRY.POINT),
+    target: Node.PARAM.NODE(Node.GEOMETRY.ANY),
+    column: Node.PARAM.STRING(),
+    method: Node.PARAM.NULLABLE(Node.PARAM.NUMBER()),
+    p1: Node.PARAM.NULLABLE(Node.PARAM.NUMBER()),
+    p2: Node.PARAM.NULLABLE(Node.PARAM.NUMBER())
+};
+
+var Enrich = Node.create(TYPE, PARAMS, {cache: true, version: 1});
+
+module.exports = Enrich;
+module.exports.TYPE = TYPE;
+module.exports.PARAMS = PARAMS;
+
+var enrichTemplate = Node.template([
+    'WITH',
+    '_source as (SELECT array_agg(the_geom) as _geo, array_agg({{=it.column}}) as _values FROM ({{=it.sourceQuery}}) _s),',
+    '_target as (SELECT *, CASE WHEN GeometryType(_t.the_geom) = \'POINT\' THEN _t.the_geom ELSE ST_Centroid(_t.the_geom) END as _geo FROM ({{=it.targetQuery}}) _t)',
+    'SELECT _target.*, CDB_SpatialInterpolation(_source._geo, _source._values, _target._geo, {{=it.method}} ,{{=it.p1}} ,{{=it.p2}} ) as {{=it.column}} FROM _target, _source',
+].join('\n'));
+
+
+Enrich.prototype.sql = function(){
+    return enrichTemplate({
+        sourceQuery: this.source.getQuery(),
+        targetQuery: this.target.getQuery(),
+        method: this.method === null ? 1 : this.method, // 0: nearest neighbor, 1: barymetric, 2: IDW
+        method: this.p1 === null ? 0 : this.p1, // IDW, number of neighbors, 0 -> unlimited
+        method: this.p2 === null ? 0 : this.p2  // IDW: decay order, 0 -> order 1
+    });
+};


lib/node/nodes/interpolation.js
```diff
@@ -0,0 +1,58 @@
...
+var Node = require('../node');
+
+var TYPE = 'enrich';
+
+var PARAMS = {
+    source: Node.PARAM.NODE(Node.GEOMETRY.POINT),
+    target: Node.PARAM.NODE(Node.GEOMETRY.ANY),
+    column: Node.PARAM.STRING(),
+    method: Node.PARAM.NULLABLE(Node.PARAM.NUMBER()),
+    p1: Node.PARAM.NULLABLE(Node.PARAM.NUMBER()),
```
dgaubert: if `p1` refers to `number of neighbours`. Could we use `numberOfNeighbours` as param name?

lib/node/nodes/interpolation.js
```diff
@@ -0,0 +1,58 @@
...
+    p2: Node.PARAM.NULLABLE(Node.PARAM.NUMBER())
```
dgaubert: Same thing here, if `p2` refers to `decay order`. Could we use `decayOrder` as param name?

lib/node/nodes/interpolation.js
```diff
@@ -0,0 +1,58 @@
...
+    '_target as (',
+    ' SELECT',
+    ' *,',
+    ' CASE WHEN GeometryType(_t.the_geom) = \'POINT\' THEN _t.the_geom',
+    ' ELSE ST_Centroid(_t.the_geom) END as _geo',
+    ' FROM ({{=it.targetQuery}}) _t',
+    ')',
+    'SELECT',
+    ' _target.*,',
+    ' CDB_SpatialInterpolation(',
```
dgaubert: I guess that `CDB_SpatialInterpolation` it's an expensive function, isn't it?. If not we should avoid CTEs and cache the analysis.

lib/node/nodes/interpolation.js
```diff
@@ -0,0 +1,58 @@
...
+    '_source._geo,',
+    ' _source._values,',
+    ' _target._geo,',
+    ' {{=it.method}} ,',
+    ' {{=it.p1}} ,',
```
dgaubert: What happen if some of NULLABLE params is effectively `null` or `undefined`?  `CDB_SpatialInterpolation` is going to work as expected?

dgaubert: This could be useful: https://github.com/CartoDB/camshaft/blob/master/lib/node/nodes/georeference-street-address.js#L64

lib/node/nodes/interpolation.js
AbelVM: Depends on the used method: https://github.com/CartoDB/crankshaft/blob/develop/doc/08_interpolation.md
Calculation time increases with method index 1>2>>0.

But sure, we should go with caching, and is set in line 16. Isn't it?

lib/node/nodes/interpolation.js
AbelVM: cool! But PostgreSQL needs the optional parameters ordered, so if you have 5 params and the last 3 are optional, you need to call the function:
1,2,3,4,5
1,2,3,4
1,2,3
With the pointed method, you might be able to send 1,2,3,5... and that is not a good idea

lib/node/nodes/interpolation.js
```diff
@@ -0,0 +1,58 @@
...
+    ' as {{=it.column}}',
+    ' FROM _target, _source'
+].join('\n'));
+
+
+Enrich.prototype.sql = function(){
+    return enrichTemplate({
+        sourceQuery: this.source.getQuery(),
+        targetQuery: this.target.getQuery(),
+        method: this.method === null ? 1 : this.method, // 0: nearest neighbor, 1: barymetric, 2: IDW
```
AbelVM: This is where I take care of nullables

dgaubert: I was worried specifically with `undefined` value but it is handled and forced to be `null` [here](https://github.com/CartoDB/camshaft/blob/master/lib/node/validator/generic.js#L8). So it's ok

dgaubert:
👍

lib/node/nodes/interpolation.js
rochoa: Why don't we enforce default values in params? And if the values are limited to just 2/3 different values, why don't we use an ENUM?

AbelVM:
Fixed:
* Using ENUM now
* Defaults defined in PARAMS

Now I like my own code much more :)

cc @rochoa

lib/node/nodes/interpolation.js
AbelVM: Done

lib/node/nodes/interpolation.js
```diff
@@ -0,0 +1,64 @@
...
+
+var PARAMS = {
+    // the scattered points dataset with real values
+    source: Node.PARAM.NODE(Node.GEOMETRY.POINT),
+    // the geometries dataset that we want to enrich with interpolated values
+    target: Node.PARAM.NODE(Node.GEOMETRY.ANY),
+    // column of values in source dataset
+    column: Node.PARAM.STRING(),
+    // 0: nearest neighbor, 1: barymetric, 2: IDW
+    method: Node.PARAM.NULLABLE(Node.PARAM.ENUM('nearest neighbor', 'barymetric','IDW'),'barymetric'),
```
rochoa: Avoid spaces in ENUM values.

lib/node/nodes/interpolation.js
```diff
@@ -0,0 +1,64 @@
...
+    // IDW, number of neighbors, 0 -> unlimited
+    numberOfNeighbours: Node.PARAM.NULLABLE(Node.PARAM.NUMBER(),0),
+    // IDW: decay order, 0 -> order 1
+    decayOrder: Node.PARAM.NULLABLE(Node.PARAM.NUMBER(),0)
```
rochoa: Is this a set of number or any number?

lib/node/nodes/interpolation.js
```diff
@@ -0,0 +1,64 @@
...
+    // the scattered points dataset with real values
+    source: Node.PARAM.NODE(Node.GEOMETRY.POINT),
+    // the geometries dataset that we want to enrich with interpolated values
+    target: Node.PARAM.NODE(Node.GEOMETRY.ANY),
+    // column of values in source dataset
+    column: Node.PARAM.STRING(),
+    // 0: nearest neighbor, 1: barymetric, 2: IDW
+    method: Node.PARAM.NULLABLE(Node.PARAM.ENUM('nearest neighbor', 'barymetric','IDW'),'barymetric'),
+    // IDW, number of neighbors, 0 -> unlimited
+    numberOfNeighbours: Node.PARAM.NULLABLE(Node.PARAM.NUMBER(),0),
```
rochoa: Use camel_case for param names.

lib/node/nodes/interpolation.js
AbelVM: Positive Integers

lib/node/nodes/interpolation.js
AbelVM: done

lib/node/nodes/interpolation.js
AbelVM: done

dgaubert: Opps, totally. I was my suggestion, my bad 😓

lib/node/nodes/interpolation.js
```diff
@@ -0,0 +1,64 @@
+'use strict';
+
+var Node = require('../node');
+
+var TYPE = 'enrich';
```
rochoa: TYPE and filename should match.

lib/node/nodes/interpolation.js
```diff
@@ -0,0 +1,64 @@
...
+    ' FROM _target, _source'
+].join('\n'));
+
+
+Enrich.prototype.sql = function(){
+    return enrichTemplate({
+        sourceQuery: this.source.getQuery(),
+        targetQuery: this.target.getQuery(),
+        method: ['nearest_neighbor', 'barymetric','IDW'].indexOf[this.method],
+        p1: this.number_of_neighbours,
```
rochoa: Do not use short names, it's free to use longer names in the templates 😛 .

AbelVM:
I'll add the tests these days and assign you the CR, @dgaubert. Sorry!  ¯\_(ツ)_/¯

lib/node/nodes/enrich.js
```diff
@@ -0,0 +1,66 @@
...
+    ' FROM _target, _source'
+].join('\n'));
+
+
+Enrich.prototype.sql = function(){
+    return enrichTemplate({
+        sourceQuery: this.source.getQuery(),
+        targetQuery: this.target.getQuery(),
+        val_column: this.val_column,
+        method: ['nearest_neighbor', 'barymetric','IDW'].indexOf[this.method],
```
dgaubert: `indexOf` is a function, don't use `[]` on it.

AbelVM: FUUUUUCK

PR #165, opened 2016-08-10
Change to use agnostic functions for isochrones and geocode-street an…

### LINKED ISSUE #164: Remove provider info and use generic functions for LDS
- [x] Changes in camshaft to not decide any default provider and delegate that into dataservices-api, which will do based on config (camshaft will only use the "generic functions" with no reference to the provider)

ethervoid:
Closes #164 

- Remove the provider parameter for geocode-street-address and trade-area analyses
- Start using provider agnostic functions from dataservices API client

### DIFF
--- a/lib/node/nodes/georeference-street-address.js
+++ b/lib/node/nodes/georeference-street-address.js
@@ -12,11 +12,7 @@ var PARAMS = {
     state: Node.PARAM.NULLABLE(Node.PARAM.STRING()),
     state_column: Node.PARAM.NULLABLE(Node.PARAM.STRING()),
     country: Node.PARAM.NULLABLE(Node.PARAM.STRING()),
-    country_column: Node.PARAM.NULLABLE(Node.PARAM.STRING()),
-    provider: Node.PARAM.NULLABLE(
-        Node.PARAM.ENUM('heremaps', 'google', 'mapzen', 'user_default'),
-        'mapzen'
-    )
+    country_column: Node.PARAM.NULLABLE(Node.PARAM.STRING())
 };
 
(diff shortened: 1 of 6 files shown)


ethervoid:
restest this please

ethervoid:
retest this please

ethervoid:
Please @dgaubert could you CR this?

// @rafatower

rafatower:
What surprises me is that there's no isodistance here (I searched the code and couldn't find it). Ok.

ethervoid:
Nope, I think the isodistance part is made with ST_Buffer instead of use cdb_isodistance function

dgaubert:
Since provider functions no longer exist, please remove its mocks in [fixtures](https://github.com/CartoDB/camshaft/tree/8f7aad4faa52759adb534b8aee5637bdc5ec7a80/test/fixtures/cdb_dataservices_client).

dgaubert:
👍

rochoa:
@rafatower Do you think it would be better to use isodistance instead of our custom st_buffer usage?

rafatower:
Indeed, I think so.

rafatower:
Pros: closer to user expectations (actual area of influence?). Cons: slower and potentially "pricey". I guess its mostly a product decision and can wait for having more stable isodistance providers.

rochoa:
Then maybe I misunderstood the problem. I thought we were talking about `buffer` analysis.

rafatower:
Probably my confusion, sorry. Then they are two different analyses: `buffer` (already implemented) and `isodistance` (to be implemented, I guess). Just ignore this thread, wrong place and timing for it.

rochoa:
Don't worry, I guess I also misguided you 😈 .

PR #170, opened 2016-08-10
Implemented analysis line-this-layer-column

dgaubert:

- [ ] [**not applicable**] Uses `{cache: true}` option when it needs full knowledge of the table it. Hints: aggregations, window functions. 
- [ ] [**not applicable**] Uses `{cache: true}` if it access external services.


dgaubert:
I've just applied needed changes

cc @rochoa

PR #171, opened 2016-08-11
Add closest features finding

AbelVM:
Answer to this feature request: https://github.com/CartoDB/crankshaft/issues/85

### DIFF
--- a/lib/node/nodes/closest.js
+++ b/lib/node/nodes/closest.js
@@ -1,0 +1,46 @@
+'use strict';
+
+var Node = require('../node');
+
+var TYPE = 'closest';
+
+var PARAMS = {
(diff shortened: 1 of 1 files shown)


dgaubert:
LGTM! 

👍

rochoa:
Hey, this is incorrect. It's using `the_geom_webmercator` instead of `the_geom`.

dgaubert:
it's using `the_geom` for the output, I figured that was enough. I'll pay more attention to that next time.

rochoa:
Don't worry. The thing is that as the output guarantees cartodb_id and the_geom, the analyses should use those.

BTW: I added another point to the template's checklist.

AbelVM:
Solved by https://github.com/CartoDB/camshaft/pull/176

PR #176, opened 2016-08-16
Fix `closest` to use the_geom

### LINKED ISSUE #174: Fix `closest` to use the_geom
cc @dgaubert

AbelVM:
Closes #174.


rochoa:
Thanks :-).

PR #178, opened 2016-08-24
Used `st_collect` only when it's needed

### LINKED ISSUE #127: centroid analysis should have option for individual geometry or entire dataset
Right now centroid find the centroid of the collected dataset (`ST_Centroid(ST_Collect(the_geom))`, but users typically want to find the centroid of each of the geometries as well (`ST_Centroid(the_geom)`). There should be an option to return one or the other of these, and should probably default to the second one.

cc @rochoa

dgaubert:
Fixes #127, validates analysis for invalid inputs, uses `st_collect` when needed.

### DIFF
--- a/lib/node/nodes/centroid.js
+++ b/lib/node/nodes/centroid.js
@@ -1,17 +1,32 @@
 'use strict';
 
 var Node = require('../node');
+var debug = require('../../util/debug')('analysis:centroid');
 
 var TYPE = 'centroid';
 
 var PARAMS = {
     source : Node.PARAM.NODE(Node.GEOMETRY.ANY),
     category_column : Node.PARAM.NULLABLE(Node.PARAM.STRING()),
-    aggregation: Node.PARAM.NULLABLE(Node.PARAM.STRING(), 'count'),
-    aggregation_column: Node.PARAM.NULLABLE(Node.PARAM.STRING())
+    aggregation: Node.PARAM.NULLABLE(Node.PARAM.ENUM('avg', 'count', 'max', 'min', 'sum'), null),
+    aggregation_column: Node.PARAM.NULLABLE(Node.PARAM.STRING()),
 };
 
-var Centroid = Node.create(TYPE, PARAMS);
+var Centroid = Node.create(TYPE, PARAMS, { version: 1,
+    beforeCreate: function(node) {
+        debug(node);
+
+        if (node.aggregation && node.aggregation !== 'count' && node.aggregation_column === null) {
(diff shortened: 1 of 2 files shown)


lib/node/nodes/centroid.js
```diff
@@ -1,17 +1,32 @@
...
+var debug = require('../../util/debug')('analysis:centroid');
 
 var TYPE = 'centroid';
 
 var PARAMS = {
     source : Node.PARAM.NODE(Node.GEOMETRY.ANY),
     category_column : Node.PARAM.NULLABLE(Node.PARAM.STRING()),
-    aggregation: Node.PARAM.NULLABLE(Node.PARAM.STRING(), 'count'),
-    aggregation_column: Node.PARAM.NULLABLE(Node.PARAM.STRING())
+    aggregation: Node.PARAM.NULLABLE(Node.PARAM.ENUM('avg', 'count', 'max', 'min', 'sum'), null),
```
rochoa: Why do we change the default `count`?

dgaubert: To make it a real `nullable`;  if `aggregation` comes with null then the `ENUM` replaces it with `count` and never is `null`

rochoa: But that's the purpose of having a default value, isn't it?

dgaubert: here we need a nullable enum, if `aggregation` comes with `null` value, we shouldn't set `count` as default, should we?

rochoa: Why not? I mean, it's nice to have the count by default, isn't it?

dgaubert: 😕  Well I'm not sure, when I used the Builder I have to check the checkbox to select aggregation function:

image

rochoa: That doesn't mean the node/analysis has to follow those defaults.

dgaubert: But if `category_column` is `null` and `st_collect` is not used, we are forcing to have a `count column` filled entirely whit `1`. right?

rochoa: If there is no category column the aggregation doesn't matter so you just don't do it. It will happen the same if there is no category column and the user provides an aggregation operation and an aggregation column, right?

dgaubert: Yes, It will. 

That's why we need to have the aggregation default to `null` to avoid to do aggregations if category column is not present.

I think, I don't get your point. 😢

dgaubert:
CR: @rochoa

PR #188, opened 2016-09-02
Fix line source to target

### LINKED ISSUE #185: line-source-to-target to closest
Using [Untitled Map 12 (on 2016-08-29 at 13.23.43).carto.zip](https://github.com/CartoDB/camshaft/files/442474/Untitled.Map.12.on.2016-08-29.at.13.23.43.carto.zip) you will find that "to closest" option only shows the closest metro relative to all stations.

I would expect to have the closest metro station for each centroid, and not just the closest, in general, one.

cc @javierarce @saleiva

dgaubert:
Fixes #185 

cc @rochoa


dgaubert:
image

Issue #206, opened 2016-10-05
Limit intermediate analysis tables size


rafatower:
Limit the size of tables created by analyses based on a percentage (10-20%) of the user's disk quota or some other sensible value.

PR #207, opened 2016-10-11
Check cached tables limits

jgoizueta:
⚠️  This requires a new version of the cartodb extension that includes a new function: https://github.com/CartoDB/cartodb-postgresql/pull/280
(so the extension must be deployed before this or all analysis will fail!!)

See #206

Each time a cache table is populted a function CDB_CheckAnalysisQuota from the cartodb extension is used to check if the limits for cached analysis tables have been exceeded, aborting the analysis in that case.

CDB_CheckAnalysisQuota raises an exception if the limits are exceeded. The message from this exception will be set to the node.

### DIFF
--- a/lib/service/database.js
+++ b/lib/service/database.js
@@ -224,6 +224,10 @@ function populateCacheTableQuery(targetTableName, outputQuery) {
     return 'INSERT INTO ' + targetTableName + ' ' + outputQuery;
 }
 
+function checkCacheTableQuery(targetTableName) {
+    return 'SELECT CDB_CheckAnalysisQuota(\'' + targetTableName + '\')';
+}
+
 function updateNodeAtAnalysisCatalogQuery(nodeIds, columns) {
     nodeIds = Array.isArray(nodeIds) ? nodeIds : [nodeIds];
     return [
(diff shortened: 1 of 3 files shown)


jgoizueta:
@rochoa can you review this?

rochoa:
👍  Thank you!

lib/service/database.js
```diff
@@ -294,7 +298,8 @@ DatabaseService.prototype.queueAnalysisOperations = function(analysis, callback)
                     id: asyncQueryId(analysis, nodeToUpdate),
                     query: transactionQuery([
                         deleteFromCacheTableQuery(targetTableName),
-                        populateCacheTableQuery(targetTableName, nodeToUpdate.sql())
+                        populateCacheTableQuery(targetTableName, nodeToUpdate.sql()),
+                        checkCacheTableQuery(targetTableName)
```
rochoa: I was thinking about this again. If we do the check after the table has been already created we are just making the analysis to fail, but we are still being able to insert into the table. Should we check before trying to insert? So if any previous table went over the limit we don't allow to create any new one.

jgoizueta: Currently, since we do the same check for every call of `checkCacheTableQuery` the actual difference would be that we would check before the first cache table and not after the last. But if we get to use the `targetTableName` argument, would we want to have the table just created or the table to be created?

To cover all bases we could have two functions and:
```
deleteFromCacheTableQuery(targetTableName),
preCheckCacheTableQuery(targetTableName),
populateCacheTableQuery(targetTableName, nodeToUpdate.sql()),
postCheckCacheTableQuery(targetTableName)
```

And do in the `preCheck...` what we are doing now.

But I'm not sure, maybe should keep this simple, just check before as you suggest to fail as soon and possible and we'll consider more complex approaches [elsewhere](https://github.com/CartoDB/camshaft/issues/209)

jgoizueta: ...on the other hand if it is the last node which creates a huge cached table, if we only check before we'll miss that (until the same or other analysis is rerun)

jgoizueta: Now that I think of it... if we only create cached tables when calling `populateCacheTableQuery` and we check after each call to it, we're already failing as soon as we can. (note that we check space allocated to cache tables) 😁

Issue #214, opened 2016-10-27
Precheck analysis and abort if they're too demanding


jgoizueta:
This is part of #209, but we're narrowing the scope of it here to just be able to prevent analysis that are clearly over the limits, before they perform any actual processing.

Specifically out of the scope of this ticket: computing global analysis requirements (rows, disk space, service quotas) and checking them against configured limits, and allowing the UI to check the requirements before executing an analysis.

PR #215, opened 2016-10-27
Add mechanism to check analysis before it's executed

jgoizueta:
See #214 

Open questions:
- We set the status to fail and keep the error message in the catalog for nodes over the limits, but this does not prevent the limits checking to be performed again if, for example, the map is reentered.
- Should we pass a `QueryRunner` instead of a `DatabaseService` to the `Requirements` constructor?
- We're using `ANALYZE` for a fast estimation of the rows in a query, and we're doing nothing if this fails (e.g. for lack of stats). We should handle that case or return to use `COUNT` (see a6aae8dce0229dc8de58a5ee5ec6f07407e22072)

Pending tasks
- [x] Determine which requirements to estimate/check for all the nodes (number columns, ...) ~~or if we instead let each node checks whatever it decides to check~~ we'll start with number of rows only
- [x] Identify which analyses we need to limit and implement `computeRequirements` for them.
- [x] Tests
- [x] Better UI integration, see https://github.com/CartoDB/cartodb/issues/10391

Some things to look at at a later time:
- Are we using node ids consistently? (e.g. to save status) [do we need to pay attention to the filters arg of id()?] we'll look at this later.
- Must we handle node filters?

### DIFF
--- a/lib/analysis.js
+++ b/lib/analysis.js
@@ -77,13 +77,13 @@ AnalysisFactory.prototype.create = function(configuration, definition, callback)
 
                 return done(null, new Analysis(rootNode));
             },
-            function analysis$register(analysis, done) {
-                databaseService.registerAnalysisInCatalog(analysis, function(err) {
+            function analysis$checkLimits(analysis, done) {
+                requirements.checkLimits(analysis, function(err) {
                     return done(err, analysis);
                 });
             },
-            function analysis$checkLimits(analysis, done) {
-                requirements.checkLimits(analysis, function(err) {
+            function analysis$register(analysis, done) {
+                databaseService.registerAnalysisInCatalog(analysis, function(err) {
                     return done(err, analysis);
                 });
             },
--- a/lib/analysis.js
+++ b/lib/analysis.js
@@ -9,6 +9,7 @@ var toposort = require('../lib/dag/toposort');
 var validator = require('../lib/dag/validator');
 
 var DatabaseService = require('./service/database');
+var Requirements = require('./service/requirements');
 
 var AnalysisLogger = require('./logging/logger');
 
@@ -60,6 +61,7 @@ AnalysisFactory.prototype.create = function(configuration, definition, callback)
         configuration.batch,
         configuration.limits
     );
+    var requirements = new Requirements(databaseService, configuration.limits);
     var logger = configuration.logger ? new AnalysisLogger(configuration.logger.stream, configuration.user) : undefined;
 
     async.waterfall(
@@ -77,6 +79,16 @@ AnalysisFactory.prototype.create = function(configuration, definition, callback)
             },
             function analysis$register(analysis, done) {
                 databaseService.registerAnalysisInCatalog(analysis, function(err) {
+                    return done(err, analysis);
+                });
+            },
+            function analysis$collectRequirements(analysis, done) {
+                requirements.computeRequirements(analysis, function(err) {
+                    return done(err, analysis);
+                });
+            },
+            function analysis$validateRequirements(analysis, done) {
+                requirements.validateRequirements(analysis, function(err) {
                     return done(err, analysis);
                 });
             },
@@ -100,10 +112,8 @@ AnalysisFactory.prototype.create = function(configuration, definition, callback)
                 if (err && err.message && err.message.match(/permission denied/i)) {
                     err = new Error('Analysis requires authentication with API key: permission denied.');
                 }
-
                 return callback(err);
             }
-
             return callback(null, analysis);
         }
     );
--- a/lib/analysis.js
+++ b/lib/analysis.js
@@ -82,13 +82,8 @@ AnalysisFactory.prototype.create = function(configuration, definition, callback)
                     return done(err, analysis);
                 });
             },
-            function analysis$collectRequirements(analysis, done) {
-                requirements.computeRequirements(analysis, function(err) {
-                    return done(err, analysis);
-                });
-            },
-            function analysis$validateRequirements(analysis, done) {
-                requirements.validateRequirements(analysis, function(err) {
+            function analysis$checkLimits(analysis, done) {
+                requirements.checkLimits(analysis, function(err) {
                     return done(err, analysis);
                 });
             },
--- a/lib/node/node.js
+++ b/lib/node/node.js
@@ -468,3 +468,31 @@ function validate(validator, params, expectedParamName) {
 
     return param;
 }
+
+Node.prototype.computeRequirements = function(databaseService, limits, callback) {
+    // TODO: by default simply compute maximum of its children
+    this.estimated_requirements = {
+      result_rows: 0
+    };
+    this.limits = {
+      result_rows: 1000000
+    };
+    return callback(null, this.requirementMessages());
+};
+
+Node.prototype.requirementMessages = function() {
+    var messages = [];
+    if (this.estimated_requirements.result_rows > this.limits.result_rows) {
+      messages.push("too many result rows");
+    }
+    return messages;
+};
+
+Node.prototype.validateRequirements = function(callback) {
+    var messages = this.requirementMessages();
+    var err;
+    if (messages.length > 0) {
+      err = new Error(messages.join("\n"));
+    }
+    callback(err);
+};
--- a/lib/node/node.js
+++ b/lib/node/node.js
@@ -470,20 +470,31 @@ function validate(validator, params, expectedParamName) {
 }
 
 Node.prototype.computeRequirements = function(databaseService, limits, callback) {
-    // TODO: by default simply compute maximum of its children
-    this.estimated_requirements = {
-      result_rows: 0
+    // By default simply compute maximum of the inputs' number of rows.
+    // TODO: if the most common multi-input analysis is some kind of join we should use
+    // the product of the input numberOfRows instead
+    var self = this;
+    var maxRows = Math.max.apply(
+        null,
+        this.inputNodes.map(function(node) {
+          return node.estimatedRequirements.numberOfRows || 0 })
+    );
+    if (maxRows < 0) {
+        maxRows = 0;
+    }
+    this.estimatedRequirements = {
+        numberOfRows: maxRows
     };
     this.limits = {
-      result_rows: 1000000
+        maximumNumberOfRows: getNodeLimit(limits, this.getType(), 'maximumNumberOfRows', 1000000)
     };
     return callback(null, this.requirementMessages());
 };
 
 Node.prototype.requirementMessages = function() {
     var messages = [];
-    if (this.estimated_requirements.result_rows > this.limits.result_rows) {
-      messages.push('too many result rows');
+    if (this.estimatedRequirements.numberOfRows > this.limits.maximumNumberOfRows) {
+        messages.push('too many result rows');
     }
     return messages;
 };
@@ -498,3 +509,12 @@ Node.prototype.validateRequirements = function(callback) {
     }
     callback(err);
 };
+
+function getNodeLimit(globalLimits, nodeType, limitName, defaultValue) {
+    var limit = (globalLimits[nodeType] == undefined) ?
+                globalLimits[limitName] :
+                globalLimits[nodeType][limitName];
+    return limit || defaultValue;
+};
+
+module.exports.getNodeLimit = getNodeLimit;
--- a/lib/node/node.js
+++ b/lib/node/node.js
@@ -475,8 +475,7 @@ Node.prototype.computeRequirements = function(databaseService, limits, callback)
     // the product of the input numberOfRows instead
     var maxRows = Math.max.apply(
         null,
-        this.inputNodes.map(function(node) {
-          return node.estimatedRequirements.numberOfRows || 0 });
+        this.inputNodes.map(function(node) { return node.estimatedRequirements.numberOfRows || 0; })
     );
     if (maxRows < 0) {
         maxRows = 0;
@@ -514,6 +513,6 @@ function getNodeLimit(globalLimits, nodeType, limitName, defaultValue) {
                 globalLimits[limitName] :
                 globalLimits[nodeType][limitName];
     return limit || defaultValue;
-};
+}
 
 module.exports.getNodeLimit = getNodeLimit;
--- a/lib/node/node.js
+++ b/lib/node/node.js
@@ -473,11 +473,10 @@ Node.prototype.computeRequirements = function(databaseService, limits, callback)
     // By default simply compute maximum of the inputs' number of rows.
     // TODO: if the most common multi-input analysis is some kind of join we should use
     // the product of the input numberOfRows instead
-    var self = this;
     var maxRows = Math.max.apply(
         null,
         this.inputNodes.map(function(node) {
-          return node.estimatedRequirements.numberOfRows || 0 })
+          return node.estimatedRequirements.numberOfRows || 0 });
     );
     if (maxRows < 0) {
         maxRows = 0;
--- a/lib/node/node.js
+++ b/lib/node/node.js
@@ -485,6 +485,8 @@ Node.prototype.computeRequirements = function(databaseService, limits, callback)
         numberOfRows: maxRows
     };
     this.limits = {
+        // Here we define a default limit for the number of rows for any kind of node which doesn't provide
+        // a computeRequirements specialization.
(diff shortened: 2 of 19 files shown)


jgoizueta:
### A note about *aliased* nodes

Let's consider this analysis definition:

```javascript
{
    id: 'a2',
    type: 'point-in-polygon',
    params: {
        points_source: {
          id: 'airbnb_rooms',
          type: 'source',
          params: {
              query: 'SELECT * FROM airbnb_rooms'
          }
        },
        polygons_source: {
            id: 'a1',
            type: 'trade-area',
            params: {
                source: {
                  id: 'a0',
                  type: 'source',
                  params: {
                      query: 'SELECT * FROM airbnb_rooms'
                  }
                },
                kind: 'car',
                time: 100,
                isolines: 1,
                dissolved: false
            }
        }
    }
};
```

The `id`s that appear in the definition are the external names (e.g. generated by the UI) of the analysis nodes, and must not be confused with the [internal ids](https://github.com/CartoDB/camshaft/blob/5113a77d6a597136d45510856cd28974e38c0ce1/lib/node/node.js#L95-L119) which are a hash of the attributes that define the node uniquely.

In this case we have two nodes, `a0` and  `airbnb_rooms` which have an identical definition and so will have the same internal id; they both represent the same node of the DAG.

If `Node` objects are created with a workflow `Factory` for this analysis it will create 4 Nodes; actually:  `new Analysis(rootNode).getNodes().length === 4`

But if we get the topologically sorted nodes we will have `new Analysis(rootNode).getNodes().length === 3`, because it includes each distinct node only once. This means that in an analysis we may have  multiple `Node` instances representing the same graph node.

In `Requirements` since we iterate through the sorted nodes, we visit all the graph nodes, but we may not visit all the `Node` objects.

If we only assign requirements to the visited Nodes, a node such as `a2` above could try to get the requirements from its "points_source" and find it has no requirements (because we visited `a0` but not `airbnb_rooms`).

This is fixed in https://github.com/CartoDB/camshaft/pull/215/commits/9808d6c25e881b77212f6a8ed007971120912543 by replicating the requirements to all Nodes having the same id.

Note that the naive solution in that commit introduces O(n<sup>2</sup>) time complexity;
this is probably not a problem since analyses should not have lots of nodes (otherwise we can to index `allNodes` by its `id()`s).

jgoizueta:
How to define per-node limits to be checked before performing an analysis.

Currently we have introduced only one limit: `maximumNumberOfRows`, the maximum number of row a node can produce. For each limit, the functions `computeRequirements` must store the value of the limit in an attribute `limits` of the node and the estimated value in   `estimatedRequirements`, for example:

```javascript
this.estimatedRequirements = { numberOfRows: 1000 };
this.limits = { maximumNumberOfRows: 1000000 };
```

In addition, the function `validateRequirements` of `Node` must compare the stored values of all the limits and pass an error to its callback if any is exceeded.

We have a default `computeRequirements` in `Node` which currently computes the maximum number of rows of all its inputs. This function can be overridden in the classes corresponding to the types of analysis that require it.

The function `computeRequirements` receives a `DatabaseService` object which can be used to access the user database. It also receives a `limits` object that when used from Windshaft-CartoDB will come from redis key `limits:analyses` in database 5.

In the limits object global limits to apply to all analysis can be passed, and also limits for specific analysis types. For example:

```javascript
limits: {
    analyses: {
         maximumNumberOfRows: 100000, // general limit
        'trade-area': {
            maximumNumberOfRows: 10000 // specifc limit
        }
    }
}
```

The compute requirements functions should use the passed limits object and have a default value for any limit. This can be fetched using the `Node.getNodeLimit` function.

jgoizueta:
Hey @dgaubert can you look at this?

cc/ @rafatower @ethervoid

lib/node/nodes/aggregate-intersection.js
```diff
@@ -77,3 +77,16 @@ var queryAggregateTemplate = Node.template([
     'WHERE ST_Intersects(_cdb_analysis_source.the_geom, _cdb_analysis_target.the_geom)',
     'GROUP BY {{=it.groupByColumns}}'
 ].join('\n'));
+
+AggregateIntersection.prototype.computeRequirements = function(databaseService, limits, callback) {
+    // we estimate the maximum possible number of rows of the result
+    var product = this.source.estimatedRequirements.numberOfRows *
```
jgoizueta: I guess this is not consider good practice, should I define an accessor and use `this.source.getEstimatedRequirements().numberOfRows` instead?
(your opinion will be welcome, @dgaubert )

dgaubert: We have several places where we don't follow `getter/setter` pattern. IMHO in Javascript is not useful, you can always get a member property directly with `.`. My advice is to avoid the `Cannot read property 'wadus' of null` using default values and so on.

lib/node/node.js
```diff
@@ -468,3 +468,56 @@ function validate(validator, params, expectedParamName) {
...
+        this.inputNodes.map(function(node) { return node.estimatedRequirements.numberOfRows || 0; })
+    );
+    if (maxRows < 0) {
+        maxRows = 0;
+    }
+    this.estimatedRequirements = {
+        numberOfRows: maxRows
+    };
+    this.limits = {
+        maximumNumberOfRows: getNodeLimit(limits, this.getType(), 'maximumNumberOfRows', 1000000)
```
jgoizueta: Since limit values should be definable in Redis, maybe we should use snake_case instead of camelCase?

dgaubert: you mean `maximumNumberOfRows`?

jgoizueta: Yes, having `maximum_number_of_rows` instead.

rafatower:
@dgaubert did you have a chance to take a further look? I'm more or less familiar with the code so I can walk you through the changes if you want.

dgaubert:
@rafatower I'm dealing with some test weird stuff in SQL-API. This afternoon let's take a look of it. Sorry!

rafatower:
sure, no problem... and thanks a lot :)

lib/analysis.js
```diff
@@ -80,6 +82,16 @@ AnalysisFactory.prototype.create = function(configuration, definition, callback)
                     return done(err, analysis);
                 });
             },
+            function analysis$collectRequirements(analysis, done) {
+                requirements.computeRequirements(analysis, function(err) {
+                    return done(err, analysis);
+                });
+            },
+            function analysis$validateRequirements(analysis, done) {
```
dgaubert: I would like take a deep think because I think we can do it during analysis creation or validation and probably before registering the analysis in catalog...

lib/service/requirements.js
```diff
@@ -0,0 +1,104 @@
...
+var Node = require('../node/node');
+var debug = require('../util/debug')('requirements');
+
+// A priori checking of the requirements/limits of an analysis
+function Requirements(databaseService, limits) {
+    this.databaseService = databaseService;
+    this.limits = limits;
+}
+
+// TODO: consider doing computation & validation in one single process
```
dgaubert: I'd implement computation & validation for each node in one single process.

lib/analysis.js
dgaubert: I'd check limits after validating analysis and before registering it. If some node reaches the limit the analysis should fail and it shouldn't be registered.

lib/service/requirements.js
```diff
@@ -0,0 +1,104 @@
...
+        },
+        callback
+    );
+};
+
+module.exports = Requirements;
+
+function replicateRequirementsToAliases(node, allNodes) {
+    var id = node.id();
+    allNodes.forEach(function(otherNode) {
```
dgaubert: Uhmm! 🤔 we need replicate in more than one node. Sorry, I didn't see it.

jgoizueta:
We have decided to release this ASAP, so I'll take care of the most relevant issues and will leave the rest for a [second iteration](https://github.com/CartoDB/camshaft/issues/220).

lib/node/node.js
```diff
@@ -468,3 +468,56 @@ function validate(validator, params, expectedParamName) {
...
+
+Node.prototype.validateRequirements = function(callback) {
+    var messages = this.requirementMessages();
+    var err;
+    if (messages.length > 0) {
+        this.status = STATUS.FAILED;
+        this.errorMessage = messages.join('\n');
+        err = new Error(this.errorMessage);
+    }
+    callback(err);
```
jgoizueta: I was making that function callable in an asynchronous way in case we needed to specialise it in cases in which limits (such as quotas) need to be obtain asyncrhonously.

But since we don't need async calls now and we'll change this design anyway I'll remove the callback.

lib/node/nodes/line-sequential.js
```diff
@@ -43,3 +43,63 @@ var routingSequentialQueryTemplate = Node.template([
...
+    '  SELECT DISTINCT {{=it.category_column}}',
+    '  FROM (',
+    '    {{=it.source}}',
+    '  ) AS _line_sequential',
+    ') AS _cdb_analysis_count_distinct'
+].join('\n'));
+
+LineSequential.prototype.computeRequirements = function(databaseService, limits, callback) {
+    var inputRows = this.source.estimatedRequirements.numberOfRows;
+    var numCategories = 1;
```
jgoizueta: Since: a) this code will be eventually replaced b) it seems to work c) the closure vars that bother you are encapsulated in a single function, I'll keep this as is, for fear of introducing new bugs. are you OK with it?

lib/node/nodes/line-sequential.js
```diff
@@ -43,3 +43,63 @@ var routingSequentialQueryTemplate = Node.template([
...
+    var self = this;
+    var rowsLimit = Node.getNodeLimit(limits, TYPE, 'maximumNumberOfRows', 1000000);
+    var pointsLimit = Node.getNodeLimit(limits, TYPE, 'maximumNumberOfPointsPerLine', 10000);
+    function storeRequirements(err) {
```
jgoizueta: For the same reasons mention above, I'd prefer to keep this as is (I consider this throw-away code have our limits working until we implement proper limit-checking)

lib/node/nodes/line-sequential.js
```diff
@@ -43,3 +43,63 @@ var routingSequentialQueryTemplate = Node.template([
...
+            maximumNumberOfPointsPerLine: pointsLimit
+        };
+        return callback(err);
+    }
+    if (this.category_column) {
+        var sql = countDistinctValues({
+            source: this.source.getQuery(),
+            category_column: this.category_column
+        });
+        sql = 'SELECT pg_sleep(30)';
```
jgoizueta: 😆  oops! a left over from one of my tests.

lib/service/database.js
```diff
@@ -48,6 +48,21 @@ DatabaseService.prototype.run = function(query, callback) {
...
 
+DatabaseService.prototype.run_writable = function(query, callback) {
+    this.queryRunner.run(query, QUERY_RUNNER_WRITE_OP, callback);
+};
+
+var TIMEOUT_ERROR_MESSAGE = new RegExp('canceling statement due to statement timeout', 'i');
+
+// check err returned by queryRunner operation for a timeout error
+DatabaseService.prototype.checkForTimeout = function(err) {
+    if (err && err.message && err.message.match(TIMEOUT_ERROR_MESSAGE)) {
```
jgoizueta: I did this to avoid returning anything other than `true` or `false` (the regular expression match). Do you think it is OK to return a truthy value different from true.

Otherwise I could:

`return !!(err && err.message && err.message.match(TIMEOUT_ERROR_MESSAGE));`

lib/node/nodes/aggregate-intersection.js
```diff
@@ -77,3 +77,16 @@ var queryAggregateTemplate = Node.template([
...
+                  this.target.estimatedRequirements.numberOfRows;
+    this.estimatedRequirements = {
+        numberOfRows: product
+    };
+    this.limits = {
+        maximumNumberOfRows: Node.getNodeLimit(limits, TYPE, 'maximumNumberOfRows', 1000000)
+    };
+    return callback(null);
```
dgaubert: Ok, but be careful with CPU intensive tasks and consider use `process.nextTick()` if required.

lib/analysis.js
jgoizueta: I'm changing all this so we get the requirements and validate in a single operations, and do it before registering, as we have talked about.

lib/service/requirements.js
jgoizueta: sure!

lib/node/nodes/line-sequential.js
dgaubert: go ahead!

dgaubert:
Let me know when you feel ready to review again!

lib/service/database.js
dgaubert: > `return !!(err && err.message && err.message.match(TIMEOUT_ERROR_MESSAGE));`

👍

jgoizueta:
Hey @dgaubert, I'll be testing this in staging, but you can already review the changes whenever it suits you

jgoizueta:
I think is ready to be deployed, @dgaubert can you take a last look at it?

The current kind of analysis are limited; the default limits can be overriden in redis:

* line-sequential: the input table cannot have more than 1M rows (even if we aggregate by some column; initially I tried to adjust for that but have [removed](https://github.com/CartoDB/camshaft/pull/215/commits/4d97ee2cdbcd26d23e5a95a112333bff79ac1344) because I think it is safer that way
* line-source-to-target: the product of the number of rows of the input tables cannot be greater than 1M.
* georeference-country: the input cannot have more than 500 rows.
* georeference-admin-region: the input cannot have more than 1000 rows.
* georeference-postal-code: if creating polygons cannot have more than 10000 input rows (otherwise the limit is 1M)

I have place some limits in other analysis not initially considered:

* buffer: if it will be dissolve, no more than 10000 input rows (otherwise 1M)
* aggregate-intersection: product of input rows <= 1M

cc @javisantana @rafatower

lib/analysis.js
```diff
@@ -80,6 +82,11 @@ AnalysisFactory.prototype.create = function(configuration, definition, callback)
                     return done(err, analysis);
                 });
             },
+            function analysis$checkLimits(analysis, done) {
```
jgoizueta: oops, I meant to do it before registration... I'm fixing it

jgoizueta:
@rochoa please take a look at this

lib/node/node.js
```diff
@@ -468,3 +471,44 @@ function validate(validator, params, expectedParamName) {
 
     return param;
 }
+
+Node.prototype.computeRequirements = function(databaseService, limits, callback) {
+    var default_limit = null; // no limit by default
```
rochoa: Why don't we use something like `Infinity` here?

lib/node/node.js
```diff
@@ -468,3 +471,44 @@ function validate(validator, params, expectedParamName) {
...
+
+    this.requirements.setMaxInputNumberOfRows(limits, default_limit);
+    return callback(null);
+};
+
+Node.prototype.requirementMessages = function() {
+    var messages = [];
+
+    // Note that any falsy value (null, 0, '', ...) means no limits:
+    if (this.requirements.getLimit('maximumNumberOfRows')) {
```
rochoa: I would hide these details in `requirements`,

lib/node/node.js
```diff
@@ -1,3 +1,4 @@
+
```
rochoa: 🙈

rochoa:
Hey, don't consider my small comments as a blocker for this.

As we will work on https://github.com/CartoDB/camshaft/issues/220, I'm pretty much OK with this first iteration. The limitations and caveats are well defined, so let's :shipit:.

BTW, good catch about `A note about aliased nodes`, I would like to address this problem at some point because it's super confusing when you need to use `getNodes` vs `getSortedNodes`.

Thanks a lot for working on this one!

javisantana:
Could we add a "limits" section in the camshaft reference so the UI can use them to show warnings? like

```json
"aggregate-intersection": {
            "params": {
                "source": {
                    "type": "node",
                    "geometry": [
                        "*"
                    ]
                    limits: {
                      max_input_rows: 100000
                    }
                },
                ...
            }

```

cc @rochoa

rochoa:
The problem with integrating directly in the reference is that this is something _dynamic_, you might want to limit to 100k rows in some cases for an analysis but it's OK in others: depending on your hardware, for instance.

PR #219, opened 2016-11-01
Changing the kmeans analysis to return text so that autostyling defau…

stuartlynn:


stuartlynn:
CR: @rochoa

rochoa:
This is incompatible with existing implementation. This should be handled at Builder level, it could detect the number of unique categories and go for histogram or category based on that.

PR #229, opened 2016-11-14
Better analysis limits checking

jgoizueta:
This takes care of  a number of issues with the analysis limits:

* Estimation of the number of rows didn't take filtering into account
* Limits could not be configured differently for some distinct analysis cases (e.g. polygon vs point postal code geocoding)

Now limits are checked before an analysis is registered and executed, and they are compared with estimations of its requirements/results, based on the `EXPLAIN` PostgreSQL command.

Note that take is care to not abort an analysis if executing EXPLAIN fails (it could be for lack of stats), ~~except if the fail is due to a timeout (which indicates the query is too complex to be successful)~~. These errors are logged (in the analysis log) with `msg`: "analysis:limits_error".

## Current Analysis limits

Limits are defined by configuration parameters. Here are the current parameters for each
type of analysis node.

### Line source to target

* `maximumNumberOfRows` limits the maximum number of rows produced. The default value is 1M for `all-to-all` and no limit for `closest`. If this limit is override in the configuration it will affect both closest and all-to-all cases.

### Line sequential

* `maximumNumberOfRows` Limits the number of input rows (points). Default value: 1M.
* `maximumNumberOfPointsPerLine` Limits the (estimated) average number of points
  per line in the result. Default value: 100000.

### Georeference country

* `maximumNumberOfRows` Limits the number of input rows. Default value: 500.

### Georeference administrative regions

* `maximumNumberOfRows` Limits the number of input rows. Default value: 1000.

### Georeference administrative regions

* `maximumNumberOfRows` Limits the number of input rows. Default value: 1000.

### Georeference postal codes

* `maximumNumberOfRows` Limits the number of input rows for point postal codes. Default value: 1M.
* `maximumNumberOfRowsPolygon` Limits the number of input rows for polygon postal codes.
  Default value: 10000.

### Buffer

* `maximumNumberOfRows` Limits the number of input rows (points). Default value: 1M.
* `maximumNumberOfRowsDissolved` Limits the number of input rows for dissolved buffers.
  Default value: 100000.

### Other analysis types

For the rest of analyses (those not defining a custom `checkLimits` method)
there's one limit currently defined:

* `maximumNumberOfRows` limits the maximum number of input rows (points).
  The default value is unlimited. So, this limit is effective only
  for those cases where a limit value is defined in the configuration.

Note that when a limit parameter is "unlimited" no access to the database is needed
to compute requirements, so this causes no overhead.

### Example

The limits can be defined in the [tiler configuration](https://github.com/CartoDB/cartodb-platform/blob/develop/cookbooks/node-windshaft/templates/default/environment.js.erb)

```javascript
{
    analysis: {
        limits: {
            "line-sequential": {
                  maximumNumberOfPointsPerLine: 100000,
                  maximumNumberOfRows: 100000
            }
        }
}
```

Notes: a limit cannot be "unset" by assigning `null` to it in the configuration; to assign `Infinity` to a limit in a JSON file use `1/0`.

Note that the limits configuration actually passed to Camshaft would be different:

```javascript
{
    limits: {
        analyses: {
            "line-sequential": {
                  maximumNumberOfPointsPerLine: 100000,
                  maximumNumberOfRows: 100000
            }
        }
}

### CODE REVIEW

**Comment by jgoizueta** [comment]
@rochoa please review this PR; here's a wrap-up for your convenience:

This changes how the estimate for the number of rows is computed; instead of all nodes storing their estimates (in `this.requirement.estimatedRequirements`) and so being able to use the computed estimates from its inputs, now only nodes that need it compute estimates autonomously, by applying EXPLAIN to its query or its inputs' queries.

The requirements logic that was added to the nodes has now be simplified, and instead of having `computeRequirements` and `requirementMessages`/`validateRequirements` in each node, now we have need only a `checkLimits` function (which estimates requirements and checks against the limits). The data "replication to aliased nodes" is no longer needed.

Since we now don't need to store limits information in each node, I've removed entirely the `requirements` object that was added to each Node instance, and the needed functions are used directly from the `node/requirements` module.

**Comment by jgoizueta** [comment]
cc @rafatower

**Comment by jgoizueta** [comment]
Hey, after some problems detected in testing I'm in the process of:

- [ ] Fix the handling of the lines all to all case
- [ ] Add logging to keep information about failed EXPLAINs (so that we can later adjust the fail behaviour)

File: lib/node/node.js [hunk 1]
```diff
@@ -472,43 +470,10 @@ function validate(validator, params, expectedParamName) {
...
-};
-
-Node.prototype.validateRequirements = function() {
-    var messages = this.requirementMessages();
-    var err = null;
-    if (messages.length > 0) {
-        err = new Error(messages.join('\n'));
-    }
-    return err;
+Node.prototype.checkLimits = function(databaseService, limits, logger, callback) {
```
rochoa: What about passing a collaborator that already is aware of `globalLimits`, `databaseService`, `logger`, so you don't have to care/understand that much about what the dependencies for each NodeRequirements method are.

jgoizueta:
@rochoa : I'm sorry for the rather extensive refactor which will make reviewing hard, but I like the code more now. I've checked this keeps working for all the cases previously tested (and fixed a small bug along).

lib/limits/estimates.js
```diff
@@ -0,0 +1,115 @@
...
+    //         var estimated_rows = null;
+    //         if (!err) {
+    //             estimated_rows = resultSet.rows[0]['QUERY PLAN'][0].Plan['Plan Rows'];
+    //         }
+    //         return done(err, estimated_rows);
+    //     });
+    //
+    // }, callback);
+    var results = {};
+    async.forEachOf(node.nodes, function(node, param, done) {
```
rochoa: As I mentioned in https://github.com/CartoDB/camshaft/pull/233#issuecomment-262262693, I would not rely on internals.

lib/limits/estimates.js
```diff
@@ -0,0 +1,115 @@
...
+
+// Estimated number of rows in the result of a node.
+// A LimitsContext object must be provided.
+// This is an asynchronous method; an error object and the number of estimated rows will
+// be passed to the callback provided.
+// This uses PostgreSQL query planner and statistics; it could fail, returning a null count
+// if no stats are available or if a timeout occurs.
+function estimatedNumberOfRows(node, context, callback) {
+    var sql = estimatedCountTemplate({
+        sourceQuery: node.sql()
```
rochoa: Why do you use `sql()` and not `getQuery()`?

lib/node/limits.js
```diff
@@ -0,0 +1,119 @@
+// Functions to apply limits to nodes
+
+'use strict';
+
+var nodeError = require('../limits/error');
+var estimates = require('../limits/estimates');
+
+function limitNumberOfRows(node, context, limit, callback) {
+    if (!limit.value || !isFinite(limit.value)) {
```
rochoa: Why do you check for `!limit.value`?

What about using `Number.isFinite`?

lib/limits/estimates.js
jgoizueta: Because we need the original query to apply EXPLAIN, not "select * from analysis_..." for cached queries.

lib/node/limits.js
jgoizueta: The `!limit.value` check is to allow for undefined limits (and ignore them).

I'll use `Number.isFinite` instead of the global one, I wasn't aware of that.

lib/limits/estimates.js
jgoizueta: Yes, I was wondering how to handle this, do you know how could I get the input nodes and the parameter name they correspond to easily? I'm trying to decide between keeping the same functionality avoiding using node.nodes and changing the functions that estimate number of input rows to require always the specific input nodes to be checked...

PR #233, opened 2016-11-21
[DO NOT MERGE YET] Avoid duplicated input nodes at node level

dgaubert:
**Note**: This PR implies some changes in UI, number of sent nodes could not be equal that received ones from Analysis API (i.e: duplicated input nodes)

CR: @rochoa

### DIFF
--- a/lib/dag/toposort.js
+++ b/lib/dag/toposort.js
@@ -86,17 +86,9 @@ function visit(node, allEdges, markedNodes, sortedNodes, parents) {
         parents[node.id()] = true;
 
         outgoingEdges.forEach(function(outgoingEdge) {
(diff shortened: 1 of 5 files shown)


rochoa:
cc @jgoizueta You might be affected by this PR as it changes inputNodes and you are using it for doing pre-flight checks.

jgoizueta:
Currently `inputNodes` is no longer used in the limits code, luckily. But the `nodes` attribute of a node is [used](https://github.com/CartoDB/camshaft/blob/03b05a8f713a1d3a13d1121d7a7cffae1430df50/lib/limits/estimates.js#L59), which I don't think is changing in this PR. Regarding that `nodes` attribute, is it an internal Node detail that should better not be used for the limits?

rochoa:
I would never rely on internals.

Issue #220, opened 2016-11-04
Analysis limits checking, second round


jgoizueta:
We have implemented some interim limits checking in #215 , but we now that implementation
to have some shortcomings (the most important being that it requires significant logic to be implemented at the node classes to estimate requirements and check the limits).

We will refactor/reimplement this feature removing that logic from the nodes: ideally, to define a new type of analysis one has only to declare its parameters and provide sql templates to generate it's queries.

Nodes will not be required to provide functions to estimate requirements or check its limits, instead they'll define their requirements in a declarative kind of way, passing data that will be used by limit-checking functions defined elsewhere.

For example, when defining a node class for a new kind of analysis we could pass limit-checking defining data to the Node factory  as follows:

```javascript
LIMITS = {
  // enable a limit checker that limits then number of estimated output rows,
  // with a default value of 10000 rows (that can be overriden in the configuration)
  max_output_rows: 100000,
  // enable a limit on the product of the estimated number of rows of each input:
  max_input_rows_product: 10000,
  // define data services quota consumptions
  dataservices: {
    // this analysis will require a geocoding quota unit per output row:
    gecoding: 1
  }
}

var MyAnalysisNode = Node.create(TYPE, PARAMS, LIMITS);
```

In the configuration passed to Camshaft (e.g. from Windshaft-Cartodb) we could provide data to override the limits for specific analysis types and also functions to check the dataservices quotas.

The limit-checking functions will need a way to obtain the estimated number of rows
of a node (and its inputs). Two possible approaches are:

* For each node, individually, execute EXPLAIN this.sql() in the user database. This does not require any implementation for each node class.
* For each node have a function that computes the estimation and stores it in the node
   (as in #215 a node could use SQL like EXPLAIN or COUNT or it could use its inputs' estimations)

We'll need to take examples from the most relevant kind of analyses and compare the kind of results we can obtain from each approach.

PR #237, opened 2016-11-29
Unify getting limit values and checking them

jgoizueta:
Now a function Checks.check handles both tasks allowing for
multiple checks (checking functions) and multiple limits per check.

See #220


lib/limits/context.js
```diff
@@ -41,7 +42,15 @@ LimitsContext.prototype.getLimit = function (nodeType, limitName, defaultValue,
         }
         limit = limits[limitName];
     }
-    return { value: limit || defaultValue, message: errorMessage };
+    if (!limit) {
+        // apply default value; a sync. function (with a node parameter) can be used
+        if (typeof defaultValue === 'function') {
```
jgoizueta: This is a remnant of a previsous iteration. I think I'll remove it, since it's not needed now

jgoizueta:
@rochoa care to take a look? Current functionality isn't change but adding or change limit checks should be easier now.

lib/limits/context.js
rochoa: If possible, I would remove it now.

lib/limits/checks.js
```diff
@@ -0,0 +1,161 @@
...
+//          name: 'maximumNumberOfRows' // (optional) name of the limit configuration parameter
+//          ...
+// The result if of the form:
+//   {  maximumNumberOfRows: {          // name of the limit argument as used by checker functions
+//          value: 1000000,             // value of the limit
+//          message: 'too many rows'    // error to be shown when limit is exceeded
+//          ...
+function limitValues(node, context, limits, callback) {
+    var values = {};
+    async.forEachOf(limits, function(limitParams, limit, done) {
```
rochoa: Why do we use async here?

lib/node/node.js
```diff
@@ -471,9 +478,8 @@ function validate(validator, params, expectedParamName) {
 }
 
 Node.prototype.checkLimits = function(context, callback) {
-    // the default is to limit number of output rows if a maximumNumberOfRows is
-    // present in the configuration (not by default)
-    var defaultLimit = null;
-    var limit = context.getLimit(this.getType(), 'maximumNumberOfRows', defaultLimit, 'too many result rows');
-    limits.limitNumberOfRows(this, context, limit, callback);
+    Checks.check(this, context, [{
```
rochoa: Do we want to have that default limit? Or should it be implicit?

I mean: could we just `return callback(null);`.

lib/node/nodes/buffer.js
```diff
@@ -122,17 +138,11 @@ Buffer.prototype.isolinesDissolvedQuery = function() {
...
-        limitName = 'maximumNumberOfRowsDissolved';
-        defaultLimit = 100000;
-    } else {
-        limitName = 'maximumNumberOfRows';
-        defaultLimit = 1000000;
-    }
-    limit = context.getLimit(TYPE, limitName, defaultLimit, 'too many source rows');
-    limits.limitInputRows(this, 'source', context, limit, callback);
+    Checks.check(this, context, [{
+        checker: Checks.limitInputRows,
```
rochoa: Isn't it a bit weird that the Checks.check function receives a reference to a function of the same module? I have the feeling this is still very _DSL_ish approach.

Why not something like:

```
this.limits.isWithinRowsLimit(this.source, limits.maximumNumberOfRows.default, callback)
```

lib/node/node.js
```diff
@@ -471,9 +478,8 @@ function validate(validator, params, expectedParamName) {
 }
 
 Node.prototype.checkLimits = function(context, callback) {
```
jgoizueta: sure

lib/node/node.js
jgoizueta: Having this here allow us to limit the number of result rows for any analysis type by setting a limit value in the configuration.

If there's no configuration value, since where have `defaultLimit` here the behaviour is equivalent to `callback(null)` (because `limitNumberOfRows` does exactly that for null/Infinite limit values).

lib/node/nodes/buffer.js
jgoizueta: The reason for the duality (`Checks.check` & `Checks.limitInputRows`) and the DSLish approach here is that `Checks.check` simplifies the processing if we want to combine multiple separate checks (e.g. check input rows and output rows), and handling for each separate check its limits values.

Then, for each check passed to `Checks.check`, the user can supply a `checker` function or use a predefined one such as `Checks.limitInputRows`here.

I see your point with the more object-oriented approach you propose, but I think we would miss the ability to easily have multiple checks. (which may not be important relly, I'm not sure... so far we're not needing it).

lib/limits/context.js
jgoizueta: I'll do

lib/limits/checks.js
jgoizueta: No need really, at some point in writing here I was expecting that getting limit values could eventually need some async operation, but now I see no reason for this. I'll remove it. My only doubt is whether to maintain this `limitValues` function as async (with callback) or not...

lib/node/nodes/buffer.js
jgoizueta: What I like about `this.limits.checkingOp(...)` vs `Checks.check(..., [{...definitions...}])` is how it simplifies some things (we don't need a check for input rows, and passing the `input` parameter since we have a check for output rows).

lib/limits/checks.js
```diff
@@ -0,0 +1,160 @@
...
+// The result if of the form:
+//   {  maximumNumberOfRows: {          // name of the limit argument as used by checker functions
+//          value: 1000000,             // value of the limit
+//          message: 'too many rows'    // error to be shown when limit is exceeded
+//          ...
+function limitValues(node, context, limits, callback) {
+    var values = {};
+    Object.keys(limits).forEach(function(limit) {
+        var limitParams = limits[limit];
+        var limitName = limit || limitParams.name;
```
jgoizueta: It should be `var limitName = limitParams.name || limit;` shouldn't it?

Issue #242, opened 2016-12-14
Integrate DS API exception safe functions


rafatower:
Replace current dataservices api client function calls with new exception safe version of functions.

See https://github.com/CartoDB/dataservices-api/issues/314

rafatower:
From @jgoizueta (thanks a lot!):

# Acceptance

Use a dedicated staging host and a user account on that host.

## 1. create some test data:

```csv 
name,address,city,county
A,Gran via 46,Madrid,Spain
B,Prieto Ureña 16,Madrid,Spain
C,Callao 4,Madrid,Spain
D,,,
E,Camino San Marcial 22,Tudela,Spain
```

Make the columns of the D record NULL! (if imported as CSV it will contain empty strings)

## 2. Check previous situation 

check that geocoding by street address fails (because of NULL address)
and no point is shown.

Creating isochrones should also fail due to the NULL geometries

## 3. Configure dedicated host with code to be tested

The dataservices API client in the user databaseg should be at the last version (with exception safe functions)
The user database (cdb_conf) should be configured to use the staging dataservices server

### setup a staging dedicated server to use the camshaft branch in the tiler:

Install camshaft branch:

* clone the camshaft repo, extract the branch
* `npm install`
* test with `PGUSER=postgres make test-all`
* create global symbolic link to this camshaft directory: `sudo npm link`

Use this camshaft in the tiler

```
cd ~/www/node-windshaft/current
npm link camshaft
sudo monit restart 'resque-imports-1'
sudo monit restart 'resque-imports-2'
sudo monit restart 'reque-imports-0'
sudo monit restart 'resque-geocodings-0'
```

## 4. Check the new code

Geocoding should now not fail, for the data above 4 points should be shown,
including the one in Tudela (after the NULL record)

Isochrones (dissolved and not) can be created wit the points.

PR #243, opened 2016-12-14
Use DS exception safe API functions

jgoizueta:
See #242
Dataservices functions are replaced by their new, exception-safe counterparts.

The functions have been replaced using this Ruby script:

```ruby
# Replace dataservices function calls by corresponding exception safe functions

DS_DIR = "~/www/dataservices-api/current"
functions_to_replace = `cat #{DS_DIR}/client/renderer/interface.yaml | grep -- '- name:' | awk '{print $3}'`.split
files_to_be_replaced = Dir['**/*.{js,sql}']

def exception_safe(f)
  "_#{f}_exception_safe"
end

files_to_be_replaced.each do |file|
  puts "Replacing in #{file}"  
  functions_to_replace.each do |fn|
    # It would be nice for this to be case insensitive, preserving case. But it isn't.
    # Note that the \b word boundaries makes this idempotent
    output = `sed -i -e "s/\\b#{fn}\\b/#{exception_safe(fn)}/g" "#{file}"`
    if !output.strip.empty?
      puts "  Replacing #{fn} :"
      puts output
    end
  end
end

# Problems: (must be solved manually)
# filenames replaced in test/setup.js (mock functions are defined in files named after the function)
```

The `test/setup.js` have been discarded.

### DIFF
--- a/lib/node/nodes/georeference-admin-region.js
+++ b/lib/node/nodes/georeference-admin-region.js
@@ -52,7 +52,7 @@ GeoreferenceAdminRegion.prototype.getFunctionParam = function (name) {
 var queryTemplate = Node.template([
     'SELECT',
     '  {{=it.columns}},',
-    '  cdb_dataservices_client.cdb_geocode_admin1_polygon(' +
+    '  cdb_dataservices_client._cdb_geocode_admin1_polygon_exception_safe(' +
     '    {{=it.geocoder_params}}',
     '  ) AS the_geom',
     'FROM ({{=it.source}}) AS _camshaft_georeference_admin_region_analysis'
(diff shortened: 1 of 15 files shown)


jgoizueta:
@rochoa this is related to https://github.com/CartoDB/dataservices-api/issues/314

We're replacing the dataservices API functions used by new (internal, undocumented) functions that do not cause exceptions on errors and return NULLs/empty sets instead without rolling back transactions.

So this changes the behaviour of the analysis in case of errors, but not the nodes' interfaces and we don't need to update the reference package (right?)

rafatower:
@rochoa just for you to know, we applied the changes we discussed, and here you have a very small doc describing the exception-safe functions, to be used from now on:

https://github.com/CartoDB/dataservices-api/blob/master/doc/internal/exception_safe.md

PR #250, opened 2017-01-17
Adds geographically weighted regression analyses (describe and predict)

ohasselblad:

- [ ] Uses `the_geom geometry(Geometry, 4326)` column for all geospatial operations.


jgoizueta:
Before merging/releasing this, please check with me @andy-esch, we have released a version of camshaft that needs some fixing before being used

andy-esch:
Thanks @jgoizueta, I'll take a look 👍

andy-esch:
We might pause on this for a while until we know it will get the Builder Analysis interface created as well. Until then we're going to create dep sql node for this instead.

PR #254, opened 2017-01-31
Speed boost for DO in UI via Camshaft OBS_GetMeta/OBS_GetData

talos:
## Summary

Update of `data-observatory-measure.js` to take advantage of `obs_getdata/obs_getmeta` to have some serious performance improvements.  This is a drop-in replacement that does not require any UI changes.

## Why we are doing this

- Performance in the existing analysis interface is very slow, on the order of 20-30QPS.
- With the deployment of several new dataservices/observatory functions (see requirements) we can use much faster methods for obtaining Observatory data

## Requirements

- https://github.com/CartoDB/dataservices-api/pull/339 must be merged and confirmed to function as expected in user accounts

## Statistics

test case | old time | new time | old qps | new qps | speedup
--- | --- | --- | --- | --- | ---
get population for 1000 US census tract polygons | 48.6s | 1.7s | 20.5 | _588_ | *28x*
get median income for 1000 US points | 47.2s | 0.2s | 21.1 | _5000_ | *236x*
get population density for 1000 US points | 47.2s | 0.2s | 21.1 | _5000_ | *236x*


talos:
@saleiva This results in some very serious performance improvements, do you think you can fit this into the roadmap sometimes soon?

Tests are failing right now, but I suspect that is at least in part because obs_getdata and obs_getmeta are not part of the dataservices-api on the test box.

saleiva:
summoing @javisantana @rochoa and @jorgesancha here... 
Do you guys think we can prioritize this before starting with the migration?

javisantana:
- Could you explain what obs_getdata/obs_getmeta is? I couldn't find any documentation. Is this some kind of batch thing?
- The first CTE could be killer, aggregate all the geometries for a table in an array is kind of hard. Did you do some test over tables with big polygons?

talos:
* Documentation is in progress, but you can find a sample here:
  - https://github.com/CartoDB/observatory-extension/blob/obs-getdata-getmeta-docs/doc/measures_functions.md#obs_getmetaextent-geometry-metadata-json
  - https://github.com/CartoDB/observatory-extension/blob/obs-getdata-getmeta-docs/doc/measures_functions.md#obs_getdatageomvals-arraygeomval-metadata-json
  - https://github.com/CartoDB/observatory-extension/blob/obs-getdata-getmeta-docs/doc/measures_functions.md#obs_getdataids-arraytext-metadata-json

* The CTE will pose problems with lots of large geometries, but also having run some tests on large geometries prior to this CTE I'm pretty sure we run into bottlenecks at other points calculations interpolations for such complex geometries before we hit bottlenecks putting everything into the array.

I was able to send 1M points through the PL/Proxy hoops, but I need to do tests to see at what point the array becomes too large with polygons.

In the interest of being able to deploy anything, I would be in favor of moving forward provide we don't actively degrade performance from what we have now.  If you look at `OBS_GetMeasure` for complex geometries (several thousand vertices per polygon) as hit in the [OBS as of 1.2.1](http://52.71.151.140/perftest/#00825e4ba115bc9d435a138ab19983c211d24f6e), we can handle ~1 per second.

screen shot 2017-02-01 at 10 52 06 am

If we can fit 100 such complex geometries into an array without hitting any errors, this is still essentially an improvement from current conditions.

The longer term solution would be to rewrite the camshaft node to chunk the requests into roughly equal-point calls to `OBS_GetData`, which would get us the best of both worlds.

I'll do some tests on how we perform with arrays of large geometries today and update.

rochoa:
> The longer term solution would be to rewrite the camshaft node to chunk the requests into roughly equal-point calls to OBS_GetData, which would get us the best of both worlds.

That assumes camshaft allows to chunk data. But I don't see that happening with the current implementation/limitations camshaft has: https://github.com/CartoDB/camshaft/blob/44bfe26100a49c394c8cca2fc51e42050ffdd15a/docs/INTERNALS.md#internals.

talos:
I'm using a liberal interpretation of the word "chunk".

I could see some fancy SQL that can do the `array_agg` of a user's rows with a sensitivity to the size of the arrays that are being aggregated.  `OBS_GetData` would then be called on each row separately, avoiding overloading the length of the array at any point.  It would all still be contained in one query, and technically all happening at once.

talos:
@rafatower @javisantana Performance tests for large geomval arrays run through `cdb_dataservices_client` to the observatory can be found [here](https://gist.github.com/talos/0e4d49619cb3a3778999dc1d5f888018).

The tests should be replicated on a production or staging server so that we fully mimic the way data is ferried.  OTOH, these tests did use PL/Proxy and are promising that we can work with several thousand large and several dozen huge geometries without having issues merely for constructing the array.  The emergent issues are down the line when we actually do interpolations and intersections, and were much worse with the old setup.

javisantana:
Thanks @talos, really useful

One interesting topic (maybe not for this time) would to to try to use TWKB. Instead of sending an array of (id, geometry) use st_astkwb, from postgis doc:

```sql
SELECT ST_AsTWKB(array_agg(geom), array_agg(gid)) FROM mytable;
                 st_astwkb
--------------------------------------------
\x040402020400000202
```

TKWB stores id inside the geometry, it uses 4x less memory and BW and is standard. The only downside is the loss of precision in some cases but not sure if the results are going to be much different. Another good thing is with tables with lot of points, doing:

```sql
SELECT ST_AsTWKB(array_agg(the_geom), array_agg(cartodb_id)) FROM (select cartodb_id, the_geom from table order by the_geom) alias
```

Will improve the payload size for those tables since the delta encoding will do its work. 

It'd improve much more if the transfer would compress the data but not sure pl/proxy is doing that (cc @rafatower )

rafatower:
@talos thanks for the report. A couple questions: what kind of failures did you get with obs_getmeta/obs_getdata? what results (or absence of results) did you get in the blank cells?

rochoa:
> One interesting topic (maybe not for this time) would to to try to use TWKB. Instead of sending an array of (id, geometry) use st_astkwb, from postgis doc:

That's interesting for some other cases where we use a similar approach, kmeans for instance.

talos:
I actually think this is an ideal use case for TWKB -- clearly this is what it was meant for.

I'll do a few tests.  I think if it doesn't cause major validity issues (a possible consequence of naive loss of precision, although conversion to TWKB and precision loss would be up to the client) it would be worse using TWKB instead of geomval from the get-go.

@rafatower the failures were unfortunately masked by the cdb_dataservices_client wrapper.  I didn't have time to re-run those on cdb_observatory directly, which would have revealed the error.  On a separate note, is there any reason why we shouldn't try to expose the internal error at least on the log, if not to the user?

The blank calls I just didn't run because the times were clearly expanding in such a manner they would take too long for me to put the document together before EOD and include them.

rafatower:
> @rafatower the failures were unfortunately masked by the cdb_dataservices_client wrapper. I didn't have time to re-run those on cdb_observatory directly, which would have revealed the error.

@talos when did you execute those tests? what user did you use? we have rollbar and the logs and we shall check them.

> On a separate note, is there any reason why we shouldn't try to expose the internal error at least on the log, if not to the user?

Security reasons: do not leak sensitive info (e.g: 3rd party api keys in a stack trace, for instance).

Have you run any specific tests to mitigate the risk of invalid geoms when simplifying?

talos:
@rafatower I have a local installation of dataservices-api I use for development, so the logs would not appear on rollbar.

We have unit tests with complicated geometries to make sure we don't generate anything invalid.  However, it can still happen as it's tricky to mitigate.  It's possible that's what happened.

rafatower:
In a local setup you can do a number of things:
- tweak postgres conf, reload/restart and check the logs dumped in `/var/log/postgresql`
- configure the logger in your dataservices-api server:
https://github.com/CartoDB/dataservices-api
```
SELECT CDB_Conf_SetConf('logger_conf', '{"geocoder_log_path": "/tmp/geocodings.log", [ "min_log_level": "[debug|info|warning|error]", "rollbar_api_key": "SERVER_SIDE_API_KEY", "log_file_path": "LOG_FILE_PATH"]}');
```
set `LOG_FILE_PATH` to something convenient to you and watch there.

talos:
Having looked into `TWKB` for a bit, it's got a lot of promise but seems to fall short when it comes to deserialization.  All PostGIS has is [ST_GeomFromTWKB](http://postgis.net/docs/manual-dev/ST_GeomFromTWKB.html), which restores a single Geometry and drops the ID information.  Even if the ID information was passed separately, there's no straightforward way to convert the passed Geometry back into an array of geometries -- `ST_Dump` is not a direct mapping, and even if one did work to stitch separate `MultiPolygon`s back together, there would be no guarantee of order so it would be impossible to ID specific geometries.

I spoke a little bit with Paul about options with TWKB.  Short of some new PostGIS code to deserialize it into (Geometry, ID) pairs, there's nothing we can do with TWKB.

Since this is a promising serialization method, we may want to think about budgeting time to adapt the existing [NodeJS deserialization](https://github.com/TWKB/twkb.js) code to C.  At that point we could consider overloading the API to support TWKB.

talos:
@javisantana Meant to mention you on the above.

@rafatower The error that occurred for `OBS_GetData/OBS_GetMeta` with 3000 complex geoms was this:

```
ERROR:  Error performing intersection: TopologyException: found non-noded intersection between LINESTRING (179.735 51.855, 179.718 51.8413) and LINESTRING (179.727 51.8464, 179.718 51.8413) at 179.71791868794949 51.841303957203266
CONTEXT:  PL/pgSQL function cdb_observatory.obs_getdata(geomval[],json,boolean) line 203 at RETURN QUERY
```

Based off of the lon/lat, it looks to me like an error with Alaska geometries, which are both a bit hairy and relatively less commonly used.  This is unrelated to this pull request and should be prioritized, diagnosed and resolved separately.  I've opened https://github.com/CartoDB/observatory-extension/issues/248 for that.

javisantana:
one question @talos, the CartoDB/observatory-extension#248 is something we had before or is it introduced with this optimization branch?

talos:
@javisantana That is a bug we had before.

javisantana:
Ok, so from my point of view we should tackle it as soon as we can (not related with this PR). We have some feedback from @saleiva about a important percent of DO analysis failing because of that.

talos:
OK!  @saleiva could you point me to more test cases of failures from this?  We've had some issues before but have cleared them out as we've pushed fixes.  New data would be helpful to narrow these problems down further.

talos:
Finally got proper access to carto-staging.com, so ran a few quick tests there.

For small geometries, it looks like our performance gains are panning out about as we'd expect -- 400 rows per second and above for polygons, as compared to ~30 before. 

For the larger geometries, the performance gains are degraded by about 40%, still leaving them much much faster than the old methods.

Above ~300 complex geometries we start seeing PL/Python errors.  Considering just 100 geometries was at the edge of what we could do before (60s), making 300 well out of reach from where we before, this is not at all a regression, but we should have our eyes on it.

Some sort of interruption kills the PL/Python wrapper:

(Create the table)

```
CREATE TABLE my_complex_geometries AS
WITH meta AS (SELECT cdb_dataservices_client.obs_getmeta(
    ST_MakeEnvelope(-179, 89, 179, -89, 4326),
    '[{"geom_id": "us.census.tiger.county"}]'
) AS meta)
SELECT row_number() over() AS cartodb_id,
       (data->0->>'value')::Geometry as the_geom
FROM cdb_dataservices_client.obs_getdata(
    ARRAY[(ST_MakeEnvelope(-179, 89, 179, -89, 4326), 1)::geomval],
    (SELECT meta FROM meta), false);
```

This query will die:

```
WITH source AS (
  SELECT ST_SetSRID(ST_Extent(the_geom), 4326) AS extent, ARRAY_AGG((the_geom, cartodb_id)::geomval) AS geomvals, COUNT(*)::int as numgeoms FROM my_complex_geometries WHERE cartodb_id <= 1000
), meta as (
  SELECT cdb_dataservices_client.obs_getmeta( extent, '[{"numer_id": "us.census.acs.B01003001", "normalization": "prenormalized"}]', 1, 1, numgeoms ) meta FROM source
)
SELECT cdb_dataservices_client.obs_getdata(geomvals, meta) FROM source, meta
```

With:

```
"cdb_dataservices_client._obs_getdata(5): [cartodb_staging_user_fef64064-3754-40b4-9032-6f7d6cd95edf_db] REMOTE ERROR: RuntimeError: Execution of function interrupted by signal"
```

Again, I don't think this should hold back this PR at all.  The boundaries of our performance are still an order of magnitude (for polygons) or two orders of magnitude (for points) ahead of where they were before.

javisantana:
Agree on moving on but it's clear to me next item is to analyze what's going on there. This is the typical situation where we pass on an issue and it hits in 2 months.

rafatower:
For the record this is one of the errors sent to rollbar from yesterday's tests:
https://rollbar.com/vizzuality/Dataservices-API/items/177/occurrences/21513022789/
```
RaiseException: cdb_dataservices_server._obs_getmeta(7): [obs_2017_01_26_ba40000ea7] REMOTE ERROR: Raster and geometry do not have the same SRID
```
which is a PostGIS error.

I'm investigating the traces a little further and will probably create a ticket with my findings.

talos:
@rafatower can spare you some investigating...

That error hits when I call `OBS_GetMeta` with `ST_Extent(the_geom)` as its first argument (across a table) forgetting to set `ST_SetSRID(..., 4326)` on the extent.

We had earlier discussed making these inputs hardier to SRID 0, perhaps defaulting to SRID 4326.  I think that would be a fine issue to raise and either resolve at the wrapper level now or fix in a future PR.  I think it's safe to assume in absence of an SRID that CARTO users are passing in 4326.

talos:
Now that the new functions are live and functional, could we get this live on staging to see how it works?

@saleiva Not sure what lane this should go into for prioritization.

rochoa:
@talos Do you mind having a look at the failing tests?

Also, we would need an issue to prioritise the release of this changes.

talos:
Yup, looking into it.  I was holding off on that because I know there are more docs in the works for running Camshaft locally, and it's impossible for me to test except by making changes & pushing to CR until that's done (https://github.com/CartoDB/cartodb-management/issues/4829).

@rochoa Into which repo should that additional issue go?

talos:
@rochoa Tests are passing, turns out while I can't do end-to-end tests I can at least solve unit test issues locally -- just had to wrap our source query and provide some missing mocks.

Would love to move this forward.  Where should I make that PR?

rochoa:
@talos The additional issue should go in this very same repo.

Yes, it's not possible to do end-to-end tests as that would require installing every single external extension. But at least you were able to discover and fix the issue with the SQL.

PR #255, opened 2017-02-10
[closest analysis] Add multiple categorized results option

AbelVM:
Adding the option to get several closest features instead of only the closest one (default to 1)

- [x] Uses `the_geom geometry(Geometry, 4326)` column for all geospatial operations.


nygeog:
This is very critical for a client :)

saleiva:
Hi guys, about this analysis.

Original plans were to use the calculate paths analysis to find nearest (and calculate lines in case you don't want routes) but I think you are talking about a different thing...

@AbelVM can we catchup on this?

nygeog:
@saleiva and @AbelVM have you spoke about this? This is a specific request from enterprise client we closed with.

rochoa:
I have been testing this one. The implementation when using categories returns multiple `category` columns, which is a problem for creating cache tables.

Apart from that, the analysis doesn't return the closest target geometries but the source geometry plus a column for the target geometry. I guess what we want is the target (closest) geometry, right?

AbelVM:
You're right. The changes in the requirements were not fully translated to a new logic. Although the results might be OK, I'll rewrite this node ASAP

rochoa:
Can you add some tests?

I've tested again and it seems the `responses` param work OK but I'm not sure about the `category` param.

The analysis doesn't return a `cartodb_id` column.

Furthermore, shouldn't the analysis return all columns from the target table?

AbelVM:
Wilco!

AbelVM:
Done!

test/acceptance/closest.js
```diff
@@ -0,0 +1,50 @@
...
+                    query: QUERY_SOURCE
+                }
+            },
+            target: {
+                type: 'source',
+                params: {
+                    query: QUERY_TARGET
+                }
+            },
+            responses: 2,
```
rochoa: Can we check the responses we get are correct?

jorgesancha:
It looks like the acceptance test is not actually testing the functionality, but rather than the results are of a type or other. We'd need to see that the tests are actually testing the functionality, that the nodes returned are specifically what is expected.

Here's an example of the type of acceptance tests we'd want to see: https://github.com/CartoDB/camshaft/blob/master/test/acceptance/node-filter.js#L42-L110

test/acceptance/closest.js
AbelVM: Added

rochoa:
@AbelVM, the tests are failing. Can you check they are testing the expected behavior?

Furthermore, we are still missing a test for the non-category version.

dgaubert:
Hey @rochoa, just for heads up!

I've tried the viewer with closest example and I got:

```
{
    "errors": ["Postgis Plugin: ERROR:  column \"category\" does not exist\nLINE 1: ..._webmercator\",1.55548e-06), 3.11097e-06)) AS geom,\"category\"...\n                                                             ^\nin executeQuery Full sql was: 'SELECT ST_AsBinary(ST_Simplify(ST_SnapToGrid(\"the_geom_webmercator\",1.55548e-06), 3.11097e-06)) AS geom,\"category\" FROM (SELECT ST_Transform(the_geom, 3857) the_geom_webmercator, cartodb_id, x, y, closest_id, closest_dist\nFROM (select * from analysis_4bd65e58e4_06ad0c6c106a83a2744a1c3af1fcbb3460a4b92b) _cdb_analysis_query) as cdbq WHERE \"the_geom_webmercator\" && ST_SetSRID('BOX3D(-20037508.3 20037508.25881302,-20037508.25881302 20037508.3)'::box3d, 3857) LIMIT 65535'\n"],
    "errors_with_context": [{
        "type": "layer",
        "message": "Postgis Plugin: ERROR:  column \"category\" does not exist\nLINE 1: ..._webmercator\",1.55548e-06), 3.11097e-06)) AS geom,\"category\"...\n                                                             ^\nin executeQuery Full sql was: 'SELECT ST_AsBinary(ST_Simplify(ST_SnapToGrid(\"the_geom_webmercator\",1.55548e-06), 3.11097e-06)) AS geom,\"category\" FROM (SELECT ST_Transform(the_geom, 3857) the_geom_webmercator, cartodb_id, x, y, closest_id, closest_dist\nFROM (select * from analysis_4bd65e58e4_06ad0c6c106a83a2744a1c3af1fcbb3460a4b92b) _cdb_analysis_query) as cdbq WHERE \"the_geom_webmercator\" && ST_SetSRID('BOX3D(-20037508.3 20037508.25881302,-20037508.25881302 20037508.3)'::box3d, 3857) LIMIT 65535'\n",
        "subtype": "query",
        "layer": {
            "id": "layer0",
            "index": 0,
            "type": "mapnik"
        }
    }]
}
```
I'll debug it a bit further.

rochoa:
False alarm 😅 . Viewer was using an old camshaft release.

PR #261, opened 2017-02-17
To make DB operations against postgres use postgres user

ethervoid:
I've been getting DB errors when try to run `npm test` so I've added the username to the DB operations to avoid the problems

// @dgaubert @rochoa


ethervoid:
I've seen in Travis we're using `PGUSER=postgres`, maybe is better to use the `test-config.js` data for it but if not I'll close this PR and update the README

PR #263, opened 2017-02-17
260 analyze after cache table creation

rafatower:
Pretty much described in the issue and in the commit messages but repeating here for convenience.

This PR ensures that ANALYZE is run along with the rest of the batch queries used for cache table creation in order to have up-to-date stats, instead of waiting for the autovacuum process to run it on the table.

The advantage of this is that stats are fresh as soon as the analysis is marked as ready, so that row count estimates can be used and the query planner can also benefit from it.

### DIFF
--- a/lib/service/database.js
+++ b/lib/service/database.js
@@ -249,6 +249,10 @@ function checkCacheTableQuery(targetTableName) {
     return 'SELECT CDB_CheckAnalysisQuota(\'' + targetTableName + '\')';
 }
 
+function analyzeTableQuery(targetTableName) {
+    return 'ANALYZE ' + targetTableName;
+}
+
 function updateNodeAtAnalysisCatalogQuery(nodeIds, columns) {
     nodeIds = Array.isArray(nodeIds) ? nodeIds : [nodeIds];
     return [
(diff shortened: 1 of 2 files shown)


lib/service/database.js
```diff
@@ -249,6 +249,10 @@ function checkCacheTableQuery(targetTableName) {
     return 'SELECT CDB_CheckAnalysisQuota(\'' + targetTableName + '\')';
 }
 
+function analyzeTableQuery(targetTableName) {
+    return 'ANALYZE ' + targetTableName;
```
dgaubert: What is it intended to accomplish?? I mean, we aren't saving `ANALYSE` output anywhere.

lib/service/database.js
```diff
@@ -321,7 +325,8 @@ DatabaseService.prototype.queueAnalysisOperations = function(analysis, callback)
                     query: annotateNodeType(nodeToUpdate) + transactionQuery([
                         deleteFromCacheTableQuery(targetTableName),
                         populateCacheTableQuery(targetTableName, nodeToUpdate.sql()),
-                        checkCacheTableQuery(targetTableName)
+                        checkCacheTableQuery(targetTableName),
+                        analyzeTableQuery(targetTableName)
```
dgaubert: What happens if `ANALYZE` fails? Should analysis creation fail too?

lib/service/database.js
dgaubert: Oh Wait!!! I've just read your first comment. Understood, sorry for the noise.

lib/service/database.js
rafatower: That's a good one. Let me think about the **possible failure modes**:
- **timeout?** I'd say the whole thing should fail. It'll take a small fraction of the analysis time (in [benchmarks](https://gist.github.com/rafatower/efbaa2924b3a1cc7b7ea0698adafe3b9), on a table of 100M rows it took 22.4 seconds). Of course there's a "corner case" in which the analysis is right below the timeout and the `ANALYZE` would make cross the timeout threshold, but...
- **permissions?** if the role has perms to write the cache table, it shall also have perms to run `ANALYZE`
- **??? any other thing?** (disk full, failover, OOM, corruptions, etc.) IMO it is best not to handle those. It is debatable but there are reasons not to do so (i.e: let the exception bubble and let the RDBMS do it's recovery stuff).

So, my vote goes for not adding exception handling code here. We can always add it later on. What's your view here?

dgaubert: My point of view is pretty similar, we could take care of it later if it's worth it. I was just wondering because I didn't know how it is feasible to fail. IMO, we should not be aware of _corner cases_ (at the moment).

dgaubert:
🇨🇭

PR #265, opened 2017-02-22
Fix DO Analysis failing in any table with an "id" column"

### LINKED ISSUE #264: DO Analysis fails in table if source table has a column called "id"
### Steps to reproduce

1. Create a blank table
2. Add a point
3. Rename the `description` column to be called `id`
3. Try to augment from DO

### What happens

Endless spinning, no data!

### .carto file

[untitled_table_41 map (on 2017-02-22 at 22.05.33).zip](https://github.com/CartoDB/camshaft/files/794950/untitled_table_41.map.on.2017-02-22.at.22.05.33.zip)

talos:
Fixes #264.

Tables with a column called `__obs_id__` will fail instead, but it's impossible to prevent some kind of overlap unless `{{=it.columns}}` is changed to specify a source table.

### DIFF
--- a/lib/node/nodes/data-observatory-measure.js
+++ b/lib/node/nodes/data-observatory-measure.js
@@ -39,7 +39,7 @@ var queryTemplate = Node.template([
     '          , null, null, numgeoms) as meta FROM _summary ',
     '  ), ',
     '_data AS ( ',
-    '  SELECT id, (data->0->>\'value\')::Numeric {{=it.final_column}} ',
+    '  SELECT id AS __obs_id__, (data->0->>\'value\')::Numeric {{=it.final_column}} ',
     '  FROM cdb_dataservices_client.OBS_GetData( ',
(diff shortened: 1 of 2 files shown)


talos:
See https://github.com/CartoDB/camshaft/issues/264 for bigger description of the issue.  This is affecting one of our major SE demos right now.

rochoa:
For this kind of regressions, I would like to see a test.

talos:
Understandable -- I  wanted a fix in ASAP since this is messing with our SEs right now, and I only found out end of day.

It should be straightforward to write a regression test when I get in today, hopefully we can still get this resolved before end of day in Madrid.  🤞

talos:
@rochoa Added test, confirmed the test fails without the code change in this PR.

rochoa:
Thanks a lot for adding the test, confirming the issue 👍 .

PR #269, opened 2017-03-07
Source node stores the affected tables

ethervoid:
We need to include the affected tables for the source nodes to be used for example to compose the cache headers.

Related cartodb/windshaft-cartodb#629

### DIFF
--- a/lib/node/node.js
+++ b/lib/node/node.js
@@ -294,14 +294,17 @@ Node.prototype.getAllInputNodes = function(filterFn) {
 
     function getAllInputNodes(node, nodeList) {
         node.getInputNodes().forEach(function(inputNode) {
-            nodeList.push(inputNode);
+            nodeList[inputNode.id()] = inputNode;
             getAllInputNodes(inputNode, nodeList);
         });
 
         return nodeList;
     }
 
-    return getAllInputNodes(this, []).filter(filterFn);
+    var inputNodes = getAllInputNodes(this, {});
+    return Object.keys(inputNodes).map(function(nodeId){
+        return inputNodes[nodeId];
+    }).filter(filterFn);
 };
 
 Node.prototype.shouldCacheQuery = function() {
--- a/lib/node/node.js
+++ b/lib/node/node.js
@@ -229,9 +229,6 @@ Node.prototype.getAffectedTables = function() {
 
 Node.prototype.addAffectedTables = function(affectedTables) {
     var tables = affectedTables;
-    if (!Array.isArray(tables))  {
-        tables = [affectedTables];
-    }
     this.affectedTables = this.affectedTables.concat(tables);
 };
 
--- a/lib/node/node.js
+++ b/lib/node/node.js
@@ -53,7 +53,8 @@ var NODE_RESERVED_KEYWORDS = {
     inputNodes: 1,
     attrs: 1,
     tags: 1,
-    setStatusFromInputNodes: 1
+    setStatusFromInputNodes: 1,
+    affectedTables: 1
 };
 
 function Node() {
@@ -82,6 +83,7 @@ function Node() {
     this.inputNodes = [];
 
     this.updatedAt = null;
+    this.affectedTables = [];
 
     this.attrs = {};
 
@@ -219,6 +221,18 @@ Node.prototype.getUpdatedAt = function() {
 
 Node.prototype.setUpdatedAt = function(updatedAt) {
     this.updatedAt = (updatedAt === null) ? null : new Date(updatedAt);
+};
+
+Node.prototype.getAffectedTables = function() {
+    return this.affectedTables;
+};
+
+Node.prototype.addAffectedTables = function(affectedTables) {
+    var tables = affectedTables;
+    if (!Array.isArray(tables))  {
+        tables = [affectedTables];
+    }
+    this.affectedTables = this.affectedTables.concat(tables);
 };
 
 Node.prototype.getLastUpdatedAtFromInputNodes = function() {
--- a/lib/node/node.js
+++ b/lib/node/node.js
@@ -291,21 +291,19 @@ Node.prototype.getInputNodes = function() {
 
 Node.prototype.getAllInputNodes = function(filterFn) {
     filterFn = filterFn || function() { return true; };
-
-    function getAllInputNodes(node, nodeList) {
-        node.getInputNodes().forEach(function(inputNode) {
-            nodeList[inputNode.id()] = inputNode;
-            getAllInputNodes(inputNode, nodeList);
-        });
-
-        return nodeList;
-    }
-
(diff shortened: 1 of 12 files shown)


lib/node/node.js
```diff
@@ -221,6 +223,18 @@ Node.prototype.setUpdatedAt = function(updatedAt) {
     this.updatedAt = (updatedAt === null) ? null : new Date(updatedAt);
 };
 
+Node.prototype.getAffectedTables = function() {
+    return this.affectedTables;
+};
+
+Node.prototype.addAffectedTables = function(affectedTables) {
+    var tables = affectedTables;
+    if (!Array.isArray(tables))  {
```
ethervoid: Cool :)

lib/node/node.js
```diff
@@ -278,6 +292,21 @@ Node.prototype.getInputNodes = function() {
...
+    function getAllInputNodes(node, nodeList) {
+        node.getInputNodes().forEach(function(inputNode) {
+            nodeList.push(inputNode);
+            getAllInputNodes(inputNode, nodeList);
+        });
+
+        return nodeList;
+    }
+
+    return getAllInputNodes(this, []).filter(filterFn);
```
ethervoid: Hum didn't think about having the same input for two different nodes, as you put in there so looks like we have to return a Set of nodes

lib/service/database.js
```diff
@@ -68,26 +68,41 @@ var affectedTableRegexCache = {
 
 DatabaseService.prototype.getLastUpdatedTimeFromAffectedTables = function(node, skip, callback) {
```
ethervoid: I thought about it but from my point of view we're returning last updated time from affected tables in the form of last_updated time and the affected table for each time entry. Don't you think?

lib/service/database.js
```diff
@@ -68,26 +68,41 @@ var affectedTableRegexCache = {
...
+        return callback(null, data);
     }
 
     var sql = node.sql()
         .replace(affectedTableRegexCache.bbox, 'ST_MakeEnvelope(0,0,0,0)')
         .replace(affectedTableRegexCache.scaleDenominator, '0')
         .replace(affectedTableRegexCache.pixelWidth, '1')
         .replace(affectedTableRegexCache.pixelHeight, '1');
-    sql = 'SELECT max(updated_at) FROM CDB_QueryTables_Updated_At($camshaft$' + sql + '$camshaft$)';
+    sql = 'SELECT max(updated_at), schema_name, table_name ' +
```
ethervoid: Hum I was doing that to remove duplicated for the same table but looking at it now seems like that is not possible, so removing group and agg function.

lib/service/database.js
```diff
@@ -111,6 +126,13 @@ DatabaseService.prototype.tableExists = function(targetTableName, callback) {
     });
 };
 
+function composeTableName(schema, table) {
```
ethervoid: Ok, good for me

lib/service/database.js
```diff
@@ -339,6 +374,10 @@ DatabaseService.prototype.queueAnalysisOperations = function(analysis, callback)
             }
         });
 
+        asyncQueries.push({
+            query: invalidateCache(Object.keys(affectedTables))
```
ethervoid: That was my point, to avoid invalidating for every node but it seems like is not a good point at all. So moving do execute the invalidation for every node

ethervoid:
Changes made, tests 🍏  in camshaft and windshaft :)

lib/service/database.js
```diff
@@ -66,28 +66,42 @@ var affectedTableRegexCache = {
...
         if (err) {
             return callback(err);
         }
 
         var defaultLastUpdatedTime = new Date('1970-01-01T00:00:00.000Z');
-        var rows = resultSet.rows || [defaultLastUpdatedTime];
-        var lastUpdatedTimeFromAffectedTables = (rows[0] && rows[0].max) || defaultLastUpdatedTime;
+        var affectedTables = [];
+        var lastUpdatedTimes = [];
+        var rows = resultSet.rows || [defaultLastUpdatedTime, null, null];
```
ethervoid: Yes because we still want to store the last updated time. I didn't want to change anything, at last null values in the affected tables are an empty array

lib/node/node.js
```diff
@@ -278,6 +289,24 @@ Node.prototype.getInputNodes = function() {
     return this.inputNodes;
 };
 
+Node.prototype.getAllInputNodes = function(filterFn) {
+    filterFn = filterFn || function() { return true; };
+
+    function getAllInputNodes(node, nodeList) {
```
ethervoid: Looks good to me, I'll change it :)

lib/service/database.js
```diff
@@ -418,6 +449,12 @@ function setCacheQueryTimeoutFromLimits(node, limits) {
     });
 }
 
+function getSourceNodes(node) {
```
dgaubert: I feel like this function is pretty useful, it could be accessible from Node, couldn't it?

lib/service/database.js
dgaubert: 👌

lib/service/database.js
```diff
@@ -66,28 +66,42 @@ var affectedTableRegexCache = {
     pixelHeight: /!pixel_height!/g
 };
 
-DatabaseService.prototype.getLastUpdatedTimeFromAffectedTables = function(node, skip, callback) {
+DatabaseService.prototype.getLastUpdatedTimeAndTableDataFromAffectedTables = function(node, skip, callback) {
```
dgaubert: Since we have `CDB_QueryTables_Updated_At` function we can use it to discover affected tables, so reuse it makes sense. But tomorrow, we will need another kind of info from `CDB_QueryTables_Updated_At` (not sure) so we will need to change again the name of this function. What about to rename it from  `getLastUpdatedTimeAndTableDataFromAffectedTables` to `getSourceMetadata`? and return an object as you already implemented, calling it `metadata` (a little less generic) instead of `data`: 

```
metadata = {
  'last_update': 'wadus-date',
  'affected_tables': 'wadus-table'
}
```
I know, it isn't perfect :man_shrugging:, does it make sense?

lib/service/database.js
ethervoid: I thought the same but talking with Raul told me that is better to avoid adding specific cases and use the `getAllInputNodes` and make your own filter.

If we think is useful enough we could add it as you said

dgaubert: If we are already aware of it, fair enough then!

lib/service/database.js
ethervoid: I don't see it like source metadata but I agree with give it a more generic name
Maybe we could call it `getTableDataFromAffectedTables` because the main purpose of this function is to retrieve data from the affected tables found.

PR #270, opened 2017-03-09
Georeference accept array of column

### LINKED ISSUE #267: Georeference should accept more than one column to compose the street address
Currently the Georeference by street address [only accepts one parameter for the street address](https://github.com/CartoDB/camshaft/blob/master/lib/node/nodes/georeference-street-address.js#L9) However, it is often the case where users data hold the street address in different columns (street type, address, number...).

The analysis should accept a template for the address param to match the UI described in CartoDB/cartodb/issues/11662.

cc @saleiva @noguerol @xavijam @rochoa

dgaubert:
Be able to accept more than one column to compose the street address

Fixes #267


rochoa:
Requirements for #267 changed, this is no longer the desired behavior.

PR #272, opened 2017-03-21
Use sumarea key for OBS_GetMeta

talos:
### Improves reliability of DO Analysis for points and sparse polygons

Mitigates https://github.com/CartoDB/observatory-extension/issues/272, whereby we did a very poor job automatically selecting a boundary when it was "sparse" (imagine a handful of census tracts scattered across the city).

It entirely solves issues we were seeing where a handful of points in a city were being augmented with high-level (county or something similar) data instead of tract level data.  This led to silly looking results where distant points got the same value from the Observatory.

Cannot be merged until [observatory-extension 1.4.0](https://github.com/CartoDB/observatory-extension/pull/277) is released.  I'll message on here when it is.

CC @makella @oriolbx as this addresses issues you were seeing.

@rochoa could you take a look so we can get this released ASAP after 1.4.0 is released?  This is also affecting some of our solutions demos right now.


talos:
## Validation steps

### From scratch

1. Create a new empty map
2. Add several points across Brooklyn, NY
3. Add an Observatory Analysis for # of subway/elevated commuters
4. Confirm that each point is getting a different number

### From prior example

1. Import this .carto file into your account: [L Train Ridership (on 2017-03-22 at 14.40.46).zip](https://github.com/CartoDB/camshaft/files/861764/L.Train.Ridership.on.2017-03-22.at.14.40.46.zip)
2. Confirm that each point is getting a different # of riders from the analysis

talos:
@rochoa The fix has been deployed on observatory-extension, so this should be OK to merge now.

makella:
@talos nice!

super important fix for telling more compelling stories at these more localized scales :)

lib/node/nodes/data-observatory-measure.js
```diff
@@ -29,14 +29,16 @@ var queryTemplate = Node.template([
     '  SELECT ST_SetSRID(ST_Extent(the_geom), 4326) extent, ',
     '         count(*)::INT numgeoms, ',
     '         ARRAY_AGG(DISTINCT ST_GeometryType(the_geom)) AS geom_types, ',
-    '         ARRAY_AGG((the_geom, cartodb_id)::geomval) geomvals ',
+    '         ARRAY_AGG((the_geom, cartodb_id)::geomval) geomvals, ',
+    '         SUM(ST_Area(the_geom)) sumarea ',
```
rochoa: Is `ST_Area` over the 4326 SRID what we want to use?

talos: Yes.  The "real" area doesn't matter, the ratio does.  We are matching it up against other areas that are computed and normalized against `4326`.  But good catch!

rochoa: Yes, but, do we expose that normalized number to the user? In that case, and if we expose the normalized value, wouldn't be better to expose a normalized value in some unit that is easier to understand for the user? 🤔

talos: No.  We use this exclusively for picking "best" geom.  We always use `::Geography` behind the scenes when calculating per sq km.

rochoa: 👍 Perfect. And sorry for the noise but I wasn't aware of that.

talos: No worries, glad you gave it a close look.  Just give me a heads up when it's on staging so I can smoke test.

PR #280, opened 2017-04-11
Deprecated SQL Function analysis

rochoa:
It allows calling a PosgreSQL function to create and populate an
analysis cache table. The function receives the name of the table
to create or populate, and the operation it has to perform, either
`create` or `populate`.

The table has to have at least:
 - A `the_geom Geometry(4326)` column.
 - A `cartodb_id numeric` column.

An example of this kind of functions can be found in the example
and in the tests included in this commit.

cc @javisantana


javisantana:
it would be possible to add a check to ensure the_geom and cartodb_id are created?

rochoa:
> it would be possible to add a check to ensure the_geom and cartodb_id are created?

I'm gonna try to add an after creating validation.

docs/deprecated-sql-function.md
```diff
@@ -0,0 +1,195 @@
...
+            SELECT array_to_string(
+              array_cat(
+                array_remove(primary_source_columns, 'the_geom'),
+                -- And we add our new `the_geom` column using the `radius` argument.
+                ARRAY['ST_Buffer(the_geom::geography, ' || radius || ')::geometry AS the_geom']
+              ),
+              ','
+            ) INTO selected_columns;
+            IF operation = 'create' THEN
+                -- In some cases you might want to drop the table.
```
javisantana: it's not clear to me when I should drop that table

docs/deprecated-sql-function.md
```diff
@@ -0,0 +1,195 @@
...
+                RAISE EXCEPTION 'Invalid method value=%', buster
+                    USING HINT = 'Valid ones are: 0=nearest neighbor, 1=barymetric, 2=IDW';
+            END IF;
+            IF operation = 'create' THEN
+                EXECUTE 'CREATE TABLE ' || table_name || ' AS ' ||
+                    -- The table will have the schema of target query but with an extra column from the source query.
+                    'SELECT _target.*, _source.' || val_column ||
+                    ' FROM (' || target_query || ') _target, (' || source_query || ') _source LIMIT 0';
+            ELSIF operation = 'populate' THEN
+                EXECUTE 'INSERT INTO ' || table_name || ' ' ||
```
javisantana: I don't know if we should try to use [this way](https://www.postgresql.org/docs/9.5/static/plpgsql-statements.html) (section 40.5.4)

```
EXECUTE command-string [ INTO [STRICT] target ] [ USING expression [, ... ] ];
```

like

```
EXECUTE 'select $1 from whatever i' into TABLE USING val_column
```

So we leave plsql to format arguments

javisantana:
Looks good to me. 

I think an exaple with pl/python would help:
- formatting long queries with plsql is a pain and would be easy with python
- having a python example would help to create those python based analysis

docs/deprecated-sql-function.md
rochoa: Agree. I would try to convert to that and see how it looks/feels.

docs/deprecated-sql-function.md
rochoa: Totally. Probably better to remove it from the documentation. I don't have a good example for it.

docs/deprecated-sql-function.md
rochoa: After reading the documentation, there is a limitation following that approach:

> Note that parameter symbols can only be used for data values — if you want to use dynamically determined table or column names, you must insert them into the command string textually.

I'm gonna try to rework the example to improve it, but for those scenarios, we need to build the string manually.

docs/deprecated-sql-function.md
```diff
@@ -0,0 +1,214 @@
...
+                    USING HINT = 'Valid ones are: 0=nearest neighbor, 1=barymetric, 2=IDW';
+            END IF;
+            IF operation = 'create' THEN
+                -- The table will have the schema of target query but with an extra column from the source query.
+                EXECUTE format(
+                    'CREATE TABLE %I AS SELECT _target.*, _source.%I FROM (%s) _target, (%s) _source LIMIT 0',
+                    table_name, val_column, target_query, source_query
+                );
+            ELSIF operation = 'populate' THEN
+                EXECUTE format($populate_query$
```
rochoa: @javisantana, do you like it more now?

rochoa:
OK. From previous suggestions, the only missing piece is the Python example.

docs/deprecated-sql-function.md
javisantana: :es:

PR #281, opened 2017-04-11
fix dupped cartodb_id and giving back the source geom when autocrossing nodes

AbelVM:
Fixes:
* Returing multiple time s the same cartodb_id when a target feature is included in the closest features of more than one source feature
* Avoid returning the source feature as target feature if we select the same node as source and target


lib/node/nodes/closest.js
```diff
@@ -35,13 +40,19 @@ var closestTemplate = Node.template([
     '        FROM target',
     '    ) AS ranked',
     '    WHERE ranking <= {{=it.responses}}',
+    '    AND not ST_equals(the_geom, source.the_geom)',
```
rochoa: Why are we excluding them in this scenario?

PR #282, opened 2017-04-11
Fix invalid date get metadata from affected tables

### LINKED ISSUE #279: Invalid date on getMetadataFromAffectedTables
Context: https://github.com/CartoDB/camshaft/blob/4ae6de5b6604d8e8d8e91ba8653bea8440356be5/lib/service/database.js#L90

If `resultSet.rows` is an empty array, it will generate a `lastUpdatedTimeFromAffectedTables` like:

```js
> new Date(Math.max.apply(null, []))
Invalid Date
```

This will fail for analyses not using user tables registered in `cdb_tablemetadata`.

cc @dgaubert

dgaubert:
Fixes #279

### DIFF
--- a/lib/service/database.js
+++ b/lib/service/database.js
@@ -87,7 +87,9 @@ DatabaseService.prototype.getMetadataFromAffectedTables = function(node, skip, c
         var defaultLastUpdatedTime = new Date('1970-01-01T00:00:00.000Z');
         var affectedTables = [];
         var lastUpdatedTimes = [];
-        var rows = resultSet.rows || [defaultLastUpdatedTime, null, null];
+        var rows = (Array.isArray(resultSet.rows) && resultSet.rows.length > 0) ?
+            resultSet.rows :
+            [defaultLastUpdatedTime, null, null];
         rows.forEach(function(row) {
             lastUpdatedTimes.push((row && row.updated_at) || defaultLastUpdatedTime);
             var schema = (row && row.schema_name) || null;
--- a/lib/service/database.js
+++ b/lib/service/database.js
@@ -89,7 +89,7 @@ DatabaseService.prototype.getMetadataFromAffectedTables = function(node, skip, c
         var lastUpdatedTimes = [];
         var rows = (Array.isArray(resultSet.rows) && resultSet.rows.length > 0) ?
             resultSet.rows :
-            [defaultLastUpdatedTime, null, null];
+            [{updated_at: defaultLastUpdatedTime, schema_name: null, table_name: null}];
         rows.forEach(function(row) {
             lastUpdatedTimes.push((row && row.updated_at) || defaultLastUpdatedTime);
             var schema = (row && row.schema_name) || null;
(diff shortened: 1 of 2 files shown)


lib/service/database.js
```diff
@@ -87,7 +87,9 @@ DatabaseService.prototype.getMetadataFromAffectedTables = function(node, skip, c
         var defaultLastUpdatedTime = new Date('1970-01-01T00:00:00.000Z');
         var affectedTables = [];
         var lastUpdatedTimes = [];
-        var rows = resultSet.rows || [defaultLastUpdatedTime, null, null];
+        var rows = (Array.isArray(resultSet.rows) && resultSet.rows.length > 0) ?
+            resultSet.rows :
+            [defaultLastUpdatedTime, null, null];
```
rochoa: I don't understand very well the fallback `rows` value.

dgaubert: 🤔

ethervoid: That's something I added but it looks wrong. It should be `[{"updated_at": defaultLastUpdatedTime, "schema_name": null, "table_name": null}]`

ethervoid: Checking below it works by coincidence 😓

dgaubert: LGTM, @rochoa do you agree?

rochoa: Isn't it better to use an empty array?

dgaubert: If we use an empty array we are falling in the same bug that we want to fix, don't we?

ethervoid: If we do that we don't enter in the [rows.forEach](https://github.com/CartoDB/camshaft/pull/282/files#diff-b62008d801fcb1406bdf7ff5e30a7a9bR93) which leads to had an empty array for `lastUpdatedTimes` and therefore we have an `Invalid Date` when we try to get the Max date as we have before

rochoa: Well, but we are faking the rows to make the forEach to work. Why don't we fix it at `lastUpdatedTimeFromAffectedTables` level?

lib/service/database.js
```diff
@@ -87,20 +87,22 @@ DatabaseService.prototype.getMetadataFromAffectedTables = function(node, skip, c
         var defaultLastUpdatedTime = new Date('1970-01-01T00:00:00.000Z');
         var affectedTables = [];
         var lastUpdatedTimes = [];
-        var rows = (Array.isArray(resultSet.rows) && resultSet.rows.length > 0) ?
-            resultSet.rows :
-            [{updated_at: defaultLastUpdatedTime, schema_name: null, table_name: null}];
+        var rows = Array.isArray(resultSet.rows) ? resultSet.rows : [];
```
ethervoid: Here could be `var rows = resulSet.rows || []` ?

dgaubert: With `var rows = resulSet.rows || []` if `resulSet.rows` is not something falsy (i.e: true) next line will try to call `forEach` and it will BOOMMM!

dgaubert: with `Array.isArray(resultSet.rows)` we are ensuring we are playing with an array.

PR #298, opened 2017-05-17
Analysis for new OBS multiple measures API

dgaubert:

- [x] Uses `the_geom geometry(Geometry, 4326)` column for all geospatial operations.
- [x] Has tests to validate the result is the expected one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
 # Changelog
 
-## 0.54.2
+## 0.55.0
 
 Released 2017-mm-dd
+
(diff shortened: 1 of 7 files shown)


rafatower:
As it is a new analysis, low risk and beautiful code, take it as fully reviewed :+1:

rafatower:
Just one thing re: a0cdcc8
All "nullable" params shall also support ['value', null, 'value'] (not just denominators). That's something I was not able to convey in the specification, sorry!

nobuti:
It works pretty good! 🙌

dgaubert:
Shall we merge this?

PR #301, opened 2017-05-23
Remove target column from sequential routing

### LINKED ISSUE #285: Remove `target_column` param for sequential routing analysis
Remove `target_column` param for sequential routing analysis and use `the_geom` as is the convention.

ethervoid:
Closes #285 

Also fixed a bug with how we get the `the_geom` column from the routing functions in dataservices

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 
-## 0.56.0
+## 0.55.1
 Released 2017-mm-dd
 
 - Removed `target column` parameter from sequenatial routing analysis
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,11 @@
 # Changelog
 
 
-## 0.55.1
+## 0.56.0
 Released 2017-mm-dd
+
+- Removed `target column` parameter from sequenatial routing analysis
+- Fixed how we get `the_geom` column in all the routing functions
 
 
 ## 0.55.0
--- a/lib/node/nodes/routing-sequential.js
+++ b/lib/node/nodes/routing-sequential.js
@@ -33,7 +33,7 @@ var routingSequentialQueryTemplate = Node.template([
     'SELECT',
     '  (route).duration,',
     '  (route).length,',
-    '  (route).the_geom',
+    '  (route).shape as the_geom',
     'FROM (',
     '  SELECT',
     '    cdb_dataservices_client._cdb_route_with_waypoints_exception_safe(',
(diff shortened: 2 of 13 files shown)


CHANGELOG.md
```diff
@@ -1,9 +1,12 @@
 # Changelog
 
 
-## 0.55.1
+## 0.56.0
```
dgaubert: Why is this a minor change?

reference/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "camshaft-reference",
-  "version": "0.33.0",
+  "version": "0.34.0",
```
dgaubert: Same here, are we adding new functionality?

lib/node/nodes/routing-sequential.js
```diff
@@ -33,11 +31,11 @@ var routingSequentialQueryTemplate = Node.template([
     'SELECT',
     '  (route).duration,',
     '  (route).length,',
-    '  (route).the_geom',
+    '  (route).shape as the_geom',
```
dgaubert: Quick question: is `shape` SRID equivalent to `the_geom`?

CHANGELOG.md
ethervoid: It's a change, we're in version 0, and is not a patch because we're breaking compatibility with possible existing tests/analysis/examples...

lib/node/nodes/routing-sequential.js
ethervoid: Yes. The returned custom type is 4326 as you can see [here](https://github.com/CartoDB/dataservices-api/blob/development/client/sql/16_custom_types.sql#L8)

CHANGELOG.md
dgaubert: Not sure what `semver` specifies in this case. We are in 0.x and we can do whatever but at the end we are fixing something not adding nothing more. As long as semver allows breaking changes in 0.x does not mean that a fix ends up in a feature, or am I missing something?

lib/node/nodes/routing-sequential.js
dgaubert: Nice! Thanks for the answer. 💋

CHANGELOG.md
ethervoid: Hum based on the changes we have two different elements here:
  - Remove the `target_column` is not a fix, is a change in the API of the function. We're removing one of the mandatory parameters of the analysis which leads to a major change. If we weren't in the 0.x version, we should release a major because this would break backward compatibility
  - Fix a bug problem on how we retrieve the geometry results from the routing functions. This change alone would be a patch

And thats why I think this is a minor change rather than a patch version

dgaubert: I agree with that but I think everything depends on the nature of this issue. If I recall correctly, the column `target_column` was introduced by me and wasn't following the convention of using `the_geom` so I made a mistake so we are fixing this.

ethervoid: Ok, then I'll use the patch version for this

ethervoid: Done

dgaubert:
I'm going to release a new version as soon as posible. @ethervoid do you think we can add this to the release?

ethervoid:
Yes, I think is pretty safe.

dgaubert:
GO AHEAD!

PR #303, opened 2017-05-23
Use left join to not lose original geometries from source

dgaubert:
After doing end to end tests in staging I realized that we were losing original geometries from source  when DO does not have data.


rafatower:
good catch! :+1:

PR #306, opened 2017-05-23
Do not pass null as string param ("null") to get metadata from DO

dgaubert:
This will fix an issue in last release. We were building DO analysis query as follows:

```
cdb_dataservices_client._OBS_GetMeta_exception_safe( extent, ('[{"numer_id": "eu.eurostat.ilc_peps11_pc_pop","denom_id": "null", "normalization": "null","geom_id": "null","numer_timespan": "2013"}]')::JSON, 1, 1, numgeoms )
```

Note that `denom_id`, `normalization` and `geom_id` have `"null"` but they must be `null`. With this hotfix query will be:

```
cdb_dataservices_client._OBS_GetMeta_exception_safe( extent, ('[{"numer_id": "eu.eurostat.ilc_peps11_pc_pop","denom_id": null, "normalization": null,"geom_id": null,"numer_timespan": "2013"}]')::JSON, 1, 1, numgeoms )
```

cc @nobuti

### DIFF
--- a/lib/node/nodes/data-observatory-multiple-measures.js
+++ b/lib/node/nodes/data-observatory-multiple-measures.js
@@ -63,10 +63,16 @@ DataObservatoryMultipleMeasures.prototype.sql = function() {
     var obsParams = this.numerators.reduce(function (params, numerator, index) {
         params.push({
             numerator: '"' + numerator + '"',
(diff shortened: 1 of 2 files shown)


dgaubert:
This issue doesn't break anything, but DO was not retrieving data if the client doesn't fill all input params.

PR #312, opened 2017-06-15
Filter NULL geometries in the geomval generation for DO analysis

### LINKED ISSUE #311: Filter NULL geometries to create the geomvals in data-observatory nodes
Related CartoDB/observatory-extension#290

We need to filter the NULL geometries in the data-observatory node query in order to avoid create empty `geomvals` that could interfere with the proper functioning of the observatory function.

So the way to deal with this in to remove NULL values [here](https://github.com/CartoDB/camshaft/blob/master/lib/node/nodes/data-observatory-measure.js#L34) and [here](https://github.com/CartoDB/camshaft/blob/master/lib/node/nodes/data-observatory-multiple-measures.js#L126)

ethervoid:
Closes #311


ethervoid:
@dgaubert this is a bit urgent to review please :). Thank you 💋

jorgesancha:
I've added the issue to CR column

lib/node/nodes/data-observatory-multiple-measures.js
```diff
@@ -123,7 +125,7 @@ var queryTemplate = Node.template([
     '_data AS ( ',
     '  SELECT id AS __obs_id__, {{=it.obsColumns}}',
     '  FROM cdb_dataservices_client._OBS_GetData_exception_safe(',
-    '    (SELECT ARRAY_AGG((the_geom, cartodb_id)::geomval) FROM _source),',
+    '    (SELECT geomvals FROM _summary),',
```
dgaubert: From my ignorance, is there a reason to change this?

ethervoid: Not really, just to improve readability and avoid possible future errors if we have to modify two where clauses instead one

ethervoid:
Thank you man :)

lib/node/nodes/data-observatory-multiple-measures.js
rafatower: Sorry for jumping here, but the summary is what is used for retrieving metadata, not the actual user's data. AFAIK just a bounding box to summarize user data, right?

rafatower: I'd appreciate an answer cause I think there's a potential problem here

ethervoid: That was the intention I guess. I moved it there to avoid duplicate the WHERE clause and made it error prone to future changes. It's the same query, more or less, we have in the old node.

I'd really appreciate if you say what potential problem we're facing to fix it and publish a new version asap. If is just about the naming I'll create an issue to have it in mind for a future version instead of publishing a new one.

rafatower: When I first saw it I thought we were just sending the summary to get the values, then I thought we were sending all the geoms to get the metadata and actually it is neither of those. Took me a third look to figure out what it was actually doing.

The potential problem is naming "summary" to something else. Confronted to the API docs confuses anyone coming after, as happened to dgaubert and myself. IMHO, naming is important enough to change it in master, even if you don't publish a new version.

rafatower: BTW in the CTE you are generating a table with the extent duplicated for every row, so the WHERE clause is not duplicated but you'll have N copies of the extent polygon in memory, one per row.

ethervoid: Ok to avoid problems, I'll move the geomvals back its place and add the where to it.

PR #317, opened 2017-08-07
Add node precheck

ethervoid:
**We need to have the _OBS_PreCheck function in the DS client part before deploying this change**

In order to be able to make pre-checks of the data that is going to be used in the analysis, I've added a `preCheckQuery` option that is executed before the main query for the analysis and inside a `READ ONLY` transaction to avoid security problems.

This PR is more to check the validity of this approach. I've tested in my local environment and it works. Analysis without precheck works fine and the cases of my precheck works too. 

Some examples

captura de pantalla 2017-08-07 13 26 07
captura de pantalla 2017-08-07 13 44 39 2

### DIFF
--- a/lib/node/nodes/data-observatory-multiple-measures.js
+++ b/lib/node/nodes/data-observatory-multiple-measures.js
@@ -49,41 +49,26 @@ var DataObservatoryMultipleMeasures = Node.create(TYPE, PARAMS, { cache: true, l
             throw new Error('The number of numerators=' + this.numerators.length +
                 ' does not match the number of numerator_timespans=' + this.numerator_timespans.length);
         }
-    },
+    }
 });
 
 module.exports = DataObservatoryMultipleMeasures;
 
 DataObservatoryMultipleMeasures.prototype._composeObsMetaParams = function() {
-    var obsParams = this.numerators.reduce(function (params, numerator, index) {
-        params.push({
-            numerator: '"' + numerator + '"',
-            normalization: '"' + this.normalizations[index]  + '"',
-            denominator: (this.denominators && this.denominators[index]) ?
-                '"' + this.denominators[index] + '"' :
-                'null',
+    var params = this.numerators.map(function (numerator, index) {
+        return {
+            numer_id: numerator,
+            denom_id: (this.denominators && this.denominators[index]) ?
+                this.denominators[index] : null,
+            normalization: this.normalizations[index],
             geom_id: (this.geom_ids && this.geom_ids[index]) ?
-                '"' + this.geom_ids[index] + '"' :
-                'null',
-            numerator_timespan: (this.numerator_timespans && this.numerator_timespans[index]) ?
-                '"' + this.numerator_timespans[index] + '"' :
-                'null'
-        });
+                this.geom_ids[index] : null,
+            numer_timespan: (this.numerator_timespans && this.numerator_timespans[index]) ?
+                this.numerator_timespans[index] : null
+        };
+    }.bind(this));
 
-        return params;
-    }.bind(this), []);
-
-    return obsParams.map(function (arg) {
-        return [
-            '{',
-                '"numer_id": ' + arg.numerator + ',',
-                '"denom_id": ' + arg.denominator + ',',
-                '"normalization": ' + arg.normalization + ',',
-                '"geom_id": ' + arg.geom_id + ',',
-                '"numer_timespan": ' + arg.numerator_timespan,
-            '}'
-        ].join('');
-    }).join(',');
+    return JSON.stringify(params);
 };
 
 DataObservatoryMultipleMeasures.prototype.sql = function() {
@@ -137,11 +122,12 @@ var queryTemplate = Node.template([
     'ON _source.cartodb_id = _data.__obs_id__'
 ].join('\n'));
 
+var preCheckQueryTemplate = Node.template([
+    'SELECT cdb_dataservices_client._OBS_PreCheck(\'{{=it.source}}\',',
+    ' \'{{=it.obsMetaParams}}\'::json)'
+].join(''));
+
 DataObservatoryMultipleMeasures.prototype.preCheckQuery = function() {
-    var preCheckQueryTemplate = Node.template([
-        'SELECT cdb_dataservices_client._OBS_PreCheck(\'{{=it.source}}\',',
-        ' \'{{=it.obsMetaParams}}\'::jsonb)'
-    ].join(''));
     var sql = preCheckQueryTemplate({
         source: this.source.getQuery(false),
         obsMetaParams: this._composeObsMetaParams()
--- a/lib/node/nodes/data-observatory-multiple-measures.js
+++ b/lib/node/nodes/data-observatory-multiple-measures.js
@@ -106,7 +106,7 @@ var queryTemplate = Node.template([
     '  SELECT',
     '    cdb_dataservices_client._OBS_GetMeta_exception_safe(',
     '      extent,',
-    '      (\'[{{=it.obsMetaParams}}]\')::JSON, 1, 1, numgeoms',
+    '      (\'{{=it.obsMetaParams}}\')::JSON, 1, 1, numgeoms',
     '    ) as meta',
     '  FROM _summary',
     '),',
--- a/lib/node/nodes/data-observatory-multiple-measures.js
+++ b/lib/node/nodes/data-observatory-multiple-measures.js
@@ -49,17 +49,12 @@ var DataObservatoryMultipleMeasures = Node.create(TYPE, PARAMS, { cache: true, l
             throw new Error('The number of numerators=' + this.numerators.length +
                 ' does not match the number of numerator_timespans=' + this.numerator_timespans.length);
         }
-    }
+    },
 });
 
 module.exports = DataObservatoryMultipleMeasures;
 
-DataObservatoryMultipleMeasures.prototype.sql = function() {
-    var finalColumns = ['_source.*']
-        .concat(this.column_names.map(function(columnName) {
-            return '_data.' + columnName;
-        })).join(', ');
-
+DataObservatoryMultipleMeasures.prototype._composeObsMetaParams = function() {
     var obsParams = this.numerators.reduce(function (params, numerator, index) {
         params.push({
             numerator: '"' + numerator + '"',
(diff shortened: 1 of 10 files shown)


lib/service/database.js
```diff
@@ -284,6 +288,12 @@ DatabaseService.prototype.queueAnalysisOperations = function(analysis, callback)
                 asyncQueries.push({
                     query: updateNodeStatusAtAnalysisCatalogQuery(nodeIds, Node.STATUS.RUNNING)
                 });
+                if (sqlWrappedNode.preCheckNodeQuery() !== undefined) {
+                    asyncQueries.push({
+                        query: readOnlyTransactionQuery([sqlWrappedNode.preCheckNodeQuery()]),
```
rochoa: Is the read only translation something we want to apply in all pre check queries?

lib/node/nodes/data-observatory-multiple-measures.js
```diff
@@ -49,17 +49,12 @@ var DataObservatoryMultipleMeasures = Node.create(TYPE, PARAMS, { cache: true, l
...
 
 module.exports = DataObservatoryMultipleMeasures;
 
-DataObservatoryMultipleMeasures.prototype.sql = function() {
-    var finalColumns = ['_source.*']
-        .concat(this.column_names.map(function(columnName) {
-            return '_data.' + columnName;
-        })).join(', ');
-
+DataObservatoryMultipleMeasures.prototype._composeObsMetaParams = function() {
```
dgaubert: I wrote this two month ago and I can't remember why I implemented it in this way. After reviewing a little, it's a really awful implementation, I guess it wasn't one of my best days.... 
 
> Can we express the reduce as a map?

Sure, indeed it's a `map` implemented as `reduce`. :man_facepalming: 

> Are we formatting the JSON within the reduce/map operations? If so, it is better to work with simple objects and format at the end with JSON.stringify, isn't it?

It is and repeating myself: I even don't know what the hell I was thinking while I was doing it.

@ethervoid 🙏  Would you mind fixing that terrible portion of code. I'd be glad to help if needed.

lib/node/nodes/data-observatory-multiple-measures.js
```diff
@@ -49,17 +49,12 @@ var DataObservatoryMultipleMeasures = Node.create(TYPE, PARAMS, { cache: true, l
             throw new Error('The number of numerators=' + node.numerators.length +
                 ' does not match the number of numerator_timespans=' + node.numerator_timespans.length);
         }
-    }
+    },
```
ethervoid: Removed. In case we want add it -> https://github.com/jshint/jshint/issues/136#issuecomment-197875841

lib/service/database.js
ethervoid: From my PoV we should. A check should not perform a write operation, so enforce it could be a life saver.

lib/node/nodes/data-observatory-multiple-measures.js
ethervoid: Sure, on it. :)

ethervoid:
Time for another round! :)

lib/service/database.js
rochoa: Then, can we add a test that verifies this behaviour?

ethervoid: [This](https://github.com/CartoDB/camshaft/pull/317/files#diff-05ff3cd3ac12bad795f26bd25a01be43R152) test should check that the pre-check query is correctly generated including the read only part.

rochoa: Fair enough. But I would prefer to test the behaviour when executing the query than the generated query.

ethervoid: So make a write query inside and check for error for example?

test/integration/analysis.js
```diff
@@ -7,6 +7,8 @@ var Analysis = require('../../lib/analysis');
 var BatchClient = require('../../lib/postgresql/batch-client');
 
 var testConfig = require('../test-config');
+var testHelper = require('../helper');
+var DataObservatoryMultipleMeasures = require('../../lib/node/nodes/data-observatory-multiple-measures');
```
ethervoid: @rochoa Don't like to include specific analysis in a generic test but I didn't found a generic analysis to deal with this. Would be a good idea to create a test analysis or something like that? or leave it like this?

rochoa: As this is a generic behaviour, you could create an ad-hoc analysis node just for testing purposes. Something in the line of https://github.com/CartoDB/camshaft/blob/4ba3e104ef7183e938c5b5f58f24e2247292abdc/test/unit/node-creation.js#L64-L91.

test/integration/analysis.js
```diff
@@ -120,6 +135,61 @@ describe('workflow', function() {
...
+                    '"normalization":"prenormalized","geom_id":"test.geoids",',
+                    '"numer_timespan":"test.timespan"}]\'::json);COMMIT;'
+                ].join('');
+                assert.equal(lastEnqueuedQuery,expectedPreCheckQuery);
+
+                done();
+            });
+        });
+
+        it.only('PRECHECK query should be inside a read-only transaction', function(done) {
```
rochoa: Don't forget to remove the `only`.

lib/node/nodes/data-observatory-multiple-measures.js
```diff
@@ -54,46 +54,35 @@ var DataObservatoryMultipleMeasures = Node.create(TYPE, PARAMS, { cache: true, l
...
+                this.denominators[index] : null,
+            normalization: this.normalizations[index],
+            geom_id: (this.geom_ids && this.geom_ids[index]) ?
+                this.geom_ids[index] : null,
+            numer_timespan: (this.numerator_timespans && this.numerator_timespans[index]) ?
+                this.numerator_timespans[index] : null
+        };
+    }.bind(this));
+
+    return JSON.stringify(params);
```
rochoa: This will generate a string version of an array that in combination with the template `(\'[{{=it.obsMetaParams}}]\')::JSON` will end having an array inside an array.

```
('[[{"numer_id":"es.ine.t1_1","denom_id":null,"normalization":"denominated","geom_id":"es.ine.the_geom","numer_timespan":"2011"}]]')::JSON
```

vs

```
('[{"numer_id":"es.ine.t1_1","denom_id":null,"normalization":"denominated","geom_id":"es.ine.the_geom","numer_timespan":"2011"}]')::JSON
```

That subtle difference will make the query to fail to find any metadata.

ethervoid: Hum there is no array [in the template](https://github.com/CartoDB/camshaft/pull/317/files/5ec93727d9a7b7d2ff97383b06b1ab5a7b588220#diff-684e239f32416fca588420dbea53cce6R127)

test/integration/analysis.js
ethervoid: Ouch!

test/integration/analysis.js
ethervoid: Great like it, thx :)

ethervoid: Yes, I thought about that but the problem is that we check for the behavior when the analysis SQL is generated by the Analysis factory. We don't only depend on the node but also in the `waterfall` operations made by the factory which only receives a node definition. In our case, we create a node but it doesn't exist in the reference. Any ideas?

lib/node/nodes/data-observatory-multiple-measures.js
rochoa: But there is one in the query for the real query.

ethervoid: Oh, didn't see it damm. sorry

ethervoid: Fixed!

test/integration/analysis.js
rochoa: My original idea was not to couple the library with the reference for this very same reason: not able to create/test a node because you need the reference, but not being able to generate the reference because you don't have the node yet.

I've just created a PR over this branch with my proposal enabling testing nodes that are not part of the library, enabling testing the logic of the Factory without having to rely on Nodes from the library.

Let me know what do you think: https://github.com/CartoDB/camshaft/pull/321.

ethervoid:
Done! :)

test/integration/analysis.js
```diff
@@ -120,6 +123,52 @@ describe('workflow', function() {
...
+                params: {
+                    table: 'airbnb_rooms'
+                }
+            };
+            testConfig.batch.inlineExecution = true;
+            Analysis.create(testConfig, definition, function(err, analysis) {
+                Factory.prototype._supportsType = supportsTypeFn;
+                Factory.prototype.getNodeClass = getNodeClassFn;
+                assert.ifError(err);
+                var catalogQuery = ['SELECT last_error_message ',
```
ethervoid: @rochoa I didn't find a way to get the last message which didn't involve to query the analysis catalog table. I'm pretty sure it exists.

The test is a bit complex this way but I think we shouldn't extract it unless we are going to make the same in other tests

rochoa: The node itself should give you access to the error message, as per https://github.com/CartoDB/camshaft/blob/25418a5148fdef7b0defdf3f5525b4774abe8c57/lib/node/node.js#L218-L220

ethervoid: Yep, but it came empty because of the inlineexecution set a true :\.

rochoa: 😈  This is a bit subtle. The first time the analysis is executed it doesn't know anything about the error because, as your test reflects, the error message is picked up from the analysis catalog. Even if the queries are executed inline, the behaviour is still async, and until the second attempt to create the analysis, the error message won't be retrieved from the catalog table.

In order to work around that situation, the test helper provides you with a `createAnalyses` method that instantiates the analysis twice in order to complete that double cycle, making it possible to pickup the error message from the catalog table.

For this test, instead of using the `Analysis.create` method from the library, I recommend you to use the test helper, as it will assist you by doing the two analysis instantiations required to complete the cycle, and it will populate the error into the node.

```diff
diff --git a/test/integration/analysis.js b/test/integration/analysis.js
index 460a222..f21d054 100644
--- a/test/integration/analysis.js
+++ b/test/integration/analysis.js
@@ -152,21 +152,16 @@ describe('workflow', function() {
                     table: 'airbnb_rooms'
                 }
             };
-            testConfig.batch.inlineExecution = true;
-            Analysis.create(testConfig, definition, function(err, analysis) {
+            var config = testConfig.create({ batch: { inlineExecution: true } });
+            testHelper.createAnalyses(definition, config, function(err, analysis) {
                 Factory.prototype._supportsType = supportsTypeFn;
                 Factory.prototype.getNodeClass = getNodeClassFn;
                 assert.ifError(err);
-                var catalogQuery = ['SELECT last_error_message ',
-                                    'FROM cdb_analysis_catalog ',
-                                    'WHERE node_id = \'' + analysis.id() + '\''].join('\n');
-                testHelper.executeQuery(catalogQuery, function(err, results) {
-                    assert.ifError(err);
-                    assert.equal(1,results.rows.length);
-                    assert.equal('cannot execute INSERT in a read-only transaction',
-                                 results.rows[0].last_error_message);
-                    done();
-                });
+                assert.equal(
+                    analysis.getRoot().getErrorMessage(),
+                    'cannot execute INSERT in a read-only transaction'
+                );
+                done();
             });
         });
```

ethervoid: :O didn't know that. Changed and working. Thank you very much for you assistance, so much to learn yet hehe

rochoa: yw 😄

PR #318, opened 2017-08-08
Bind node itself in beforeCreate function for better semantics

rochoa:
It's easier to understand what you are affecting when you work
with `this` instead with a pass by reference `node` variable.

This will align with other hooks like one for being able to execute
an arbitrary query before creating the node to ensure/validate some
metadata is present in the created node.

### DIFF
--- a/lib/node/node.js
+++ b/lib/node/node.js
@@ -449,7 +449,7 @@ module.exports.create = function createNode(type, expectedParams, options) {
         // `beforeCreate` allows to change/modify the node before it's actually created/returned
         // if it throws an error it will fail to create the node.
         if (options.beforeCreate) {
-            options.beforeCreate(this);
+            options.beforeCreate.bind(this)();
         }
         this.validators = Array.isArray(options.validators) ? options.validators : [];
     }
--- a/lib/node/nodes/bounding-box.js
+++ b/lib/node/nodes/bounding-box.js
@@ -11,8 +11,8 @@ var PARAMS = {
 };
 
 var BoundingBox = Node.create(TYPE, PARAMS, { version: 1,
-    beforeCreate: function(node) {
-        if (node.aggregation !== 'count' && node.aggregation_column === null) {
+    beforeCreate: function() {
+        if (this.aggregation !== 'count' && this.aggregation_column === null) {
             throw new Error('Param `aggregation` != "count" requires an existent `aggregation_column` column');
         }
     }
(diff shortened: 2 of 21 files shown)


dgaubert:
Apart from aligning with `preCheck` hook. Is your idea to move `beforeCreate` hook from options to a object method for each node's subclasses or are we going to keep both approach?

rochoa:
Interesting question!

Could you outline how that code could be? I mean, how would you define a node with that _object method_?

dgaubert:
Something like this: https://github.com/CartoDB/camshaft/compare/before-create-bind...before-create-as-object-method

That leads to rely more on inheritance which is something that I'm not sure If you want to avoid. 😈

rochoa:
Yep, that's an option we have.

Actually, when I was doing the changes for this PR, for a moment I thought about doing:

```
this.beforeCreate = options.beforeCreate.bind(this);
this.beforeCreate();
```

So that would be compatible with your approach, allowing to override it in the prototype. What do you think about doing that change? It would be something like:

```diff
diff --git a/lib/node/node.js b/lib/node/node.js
index 42112be..e3a5a62 100644
--- a/lib/node/node.js
+++ b/lib/node/node.js
@@ -367,6 +367,9 @@ Node.prototype.toDot = function() {
     };
 };

+Node.prototype.beforeCreate = function() {
+};
+
 module.exports.PARAM = {
     NODE: function() {
         var args = (arguments.length === 1 ? [arguments[0]] : Array.apply(null, arguments));
@@ -449,8 +452,9 @@ module.exports.create = function createNode(type, expectedParams, options) {
         // `beforeCreate` allows to change/modify the node before it's actually created/returned
         // if it throws an error it will fail to create the node.
         if (options.beforeCreate) {
-            options.beforeCreate.bind(this)();
+            this.beforeCreate = options.beforeCreate.bind(this);
         }
+        this.beforeCreate();
         this.validators = Array.isArray(options.validators) ? options.validators : [];
     }
     util.inherits(BaseNode, Node);
```

dgaubert:
Pretty nice! 💪 

Having a default `beforeCreate` that could be overridden by subclassing and/or by options sounds really good. 

Being a PITA, the following question is: are we going to need such mechanism in the near future? ;)

rochoa:
I don't think so. All the existing analyses use the `Node.create` for defining the classes. So I would keep it simple. Thoughts?

dgaubert:
Neither do I. Let's keep it simple.

More than once I had a thought if a classic design pattern could be applicable here. No sure if it's worth refactoring.

PR #321, opened 2017-08-17
Factory can accept an optional map for node types/implementations

rochoa:
This allows, as per included factory integration suite, testing nodes that do not exist in the library but in the tests themselves.

This can be useful to validate Factory logic.

### DIFF
--- a/lib/workflow/factory.js
+++ b/lib/workflow/factory.js
@@ -5,15 +5,15 @@ var Node = require('../node/node');
 var async = require('async');
 var debug = require('../util/debug')('factory');
 
-var typeNodeMap = nodes.reduce(function(typeNodeMap, node) {
+var DEFAULT_TYPE_NODE_MAP = nodes.reduce(function(typeNodeMap, node) {
     typeNodeMap[node.TYPE] = node;
     return typeNodeMap;
 }, {});
 
-function Factory(user, databaseService) {
+function Factory(user, databaseService, typeNodeMap) {
     this.user = user;
     this.databaseService = databaseService;
-    this.typeNodeMap = typeNodeMap;
+    this.typeNodeMap = typeNodeMap || DEFAULT_TYPE_NODE_MAP;
 }
 
 module.exports = Factory;
--- a/test/integration/factory.js
+++ b/test/integration/factory.js
@@ -1,0 +1,54 @@
+'use strict';
+
+var assert = require('assert');
+
+var Node = require('../../lib/node/node');
+var DatabaseService = require('../../lib/service/database');
+var Factory = require('../../lib/workflow/factory');
+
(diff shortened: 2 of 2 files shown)


lib/workflow/factory.js
```diff
@@ -5,15 +5,15 @@ var Node = require('../node/node');
 var async = require('async');
 var debug = require('../util/debug')('factory');
 
-var typeNodeMap = nodes.reduce(function(typeNodeMap, node) {
+var DEFAULT_TYPE_NODE_MAP = nodes.reduce(function(typeNodeMap, node) {
```
ethervoid: One question. Why in the `var nodes = ...` we don't require `../node/nodes` ?

lib/workflow/factory.js
```diff
@@ -5,15 +5,15 @@ var Node = require('../node/node');
...
     typeNodeMap[node.TYPE] = node;
     return typeNodeMap;
 }, {});
 
-function Factory(user, databaseService) {
+function Factory(user, databaseService, typeNodeMap) {
     this.user = user;
     this.databaseService = databaseService;
-    this.typeNodeMap = typeNodeMap;
+    this.typeNodeMap = typeNodeMap || DEFAULT_TYPE_NODE_MAP;
```
ethervoid: Don't know if it's better to assign or merge the new `typeNodeMap` with the default one.

On one hand, we are able to reuse the Factory in the tests and we're able to make tests for test nodes and existing nodes, on the other hand, this could be a bit messy.

What do you think?

lib/workflow/factory.js
rochoa: Two things:
 - We want to have them indexed by type.
 - The `../node/nodes` directory doesn't have an `index.js`, so it won't provide us with the nodes list.

lib/workflow/factory.js
rochoa: I prefer to keep it simple, you can always do the merging phase outside.

lib/workflow/factory.js
ethervoid: Great, didn't know about the index.js :D

lib/workflow/factory.js
ethervoid: Perfect then!

rochoa:
Should we merge this into your branch (this PR)? Or do you prefer I merge it to master (new PR) and you pull it from there?

ethervoid:
I'd merge it into master and I'll rebase this changes into my branch

ethervoid:
but I see you have pick this from my branch so merge it into my branch

ethervoid:
Thank you very much for taking time to help me with this :)

rochoa:
Whatever you prefer, I could open a new PR to master.

PR #323, opened 2017-08-17
Allow executing an async function before the node is created

### LINKED ISSUE #316: Allow executing arbitrary SQL for analysis preparation
Working on https://github.com/CartoDB/bigmetadata/issues/227, I need the DO analysis to be able to return an schema that is dependent on the query itself, i.e: depending on the columns requested to `OBS_GetData`, the result will have different column types (as measures might be numeric or texts).

Due to how PostgreSQL works, the query must be able to know the schema before being executed (i.e: a query cannot return different schemas based on parameters), so I need to execute a call to `OBS_GetMeta` before running the analysis, in order to determine the schema, and then use that to craft the final query.

In the current camshaft, there is no way to execute such a query, since the analysis nodes do not have access to the psql connection. After talking with @rochoa, our proposal is to add a way to do this, adding a new hook to the analysis nodes, so you can write something similar to:
```js
beforeExecute: function(node, queryRunner, done) {
  queryRunner.execute('SELECT...', (err, result) => {
    if (err) {
      done(new Error('wadus'));
    } else {
      node.meta = result;
      done();
    }
  }
}
```

This can be used to solve our current problem with the DO node, but it should also be flexible enough to implement pre-checks for different analysis or, in general, any operations that need to be done by a node that fall outside the execution of a single query.

cc @ethervoid This might be useful to call the pre-check code.

rochoa:
Allow executing an async function before the node is created. This allows preparing the node using metadata from the database.

For instance, in DO analysis we will retrieve the column types for the numerators.

This closes #316.

### DIFF
--- a/lib/node/node.js
+++ b/lib/node/node.js
@@ -367,6 +367,14 @@ Node.prototype.toDot = function() {
     };
 };
 
+// ----- hooks ------
+
+Node.prototype.beforeCreate = function() {};
+
+Node.prototype.beforeCreateAsync = function(databaseService, callback) {
+    return callback(null);
+};
+
 module.exports.PARAM = {
     NODE: function() {
         var args = (arguments.length === 1 ? [arguments[0]] : Array.apply(null, arguments));
@@ -449,7 +457,11 @@ module.exports.create = function createNode(type, expectedParams, options) {
         // `beforeCreate` allows to change/modify the node before it's actually created/returned
         // if it throws an error it will fail to create the node.
         if (options.beforeCreate) {
-            options.beforeCreate.bind(this)();
+            this.beforeCreate = options.beforeCreate.bind(this);
+        }
+        this.beforeCreate();
+        if (options.beforeCreateAsync) {
+            this.beforeCreateAsync = options.beforeCreateAsync.bind(this);
         }
         this.validators = Array.isArray(options.validators) ? options.validators : [];
     }
--- a/lib/node/nodes/data-observatory-multiple-measures.js
+++ b/lib/node/nodes/data-observatory-multiple-measures.js
@@ -49,6 +49,36 @@ var DataObservatoryMultipleMeasures = Node.create(TYPE, PARAMS, { cache: true, l
             throw new Error('The number of numerators=' + this.numerators.length +
                 ' does not match the number of numerator_timespans=' + this.numerator_timespans.length);
         }
+
+        this.columnTypes = this.numerators.reduce((columnTypes, numeratorId, index) => {
+            columnTypes[numeratorId] = {
+                name: this.column_names[index],
+                type: 'Numeric'
+            };
+            return columnTypes;
+        }, {});
+    },
+    beforeCreateAsync: function(databaseService, callback) {
+        const sql = columnTypesQueryTemplate({
+            source: this.source.getQuery(false),
+            obsMetaParams: this._composeObsMetaParams()
+        });
+        databaseService.run(sql, (err, result) => {
+            if (err) {
+                return callback(err);
+            }
+
+            const measures = (Array.isArray(result.rows) && result.rows[0]) || {};
+            measures.meta = measures.meta || [];
+            measures.meta.forEach(measureMetadata => {
+                if (this.columnTypes.hasOwnProperty(measureMetadata.numer_id)) {
+                    const type = (measureMetadata.numer_type.toLowerCase() === 'text') ? 'Text' : 'Numeric';
+                    this.columnTypes[measureMetadata.numer_id].type = type;
+                }
+            });
+
+            return callback(null);
+        });
     }
 });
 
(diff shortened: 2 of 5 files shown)


rochoa:
cc @javitonino

lib/node/node.js
```diff
@@ -449,7 +457,11 @@ module.exports.create = function createNode(type, expectedParams, options) {
         // `beforeCreate` allows to change/modify the node before it's actually created/returned
         // if it throws an error it will fail to create the node.
         if (options.beforeCreate) {
-            options.beforeCreate.bind(this)();
+            this.beforeCreate = options.beforeCreate.bind(this);
```
ethervoid: Why don't we make everything async?

I suppose is because we don't want to change all the `beforeCreate` methods but to have two methods is a bit confusing and dangerous also as @dgaubert always say we don't want to [release the zalgo](http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony)

What do you think? We can leave it this way and create a ticket to change it in the future?

lib/node/nodes/data-observatory-multiple-measures.js
```diff
@@ -49,6 +49,36 @@ var DataObservatoryMultipleMeasures = Node.create(TYPE, PARAMS, { cache: true, l
             throw new Error('The number of numerators=' + this.numerators.length +
                 ' does not match the number of numerator_timespans=' + this.numerator_timespans.length);
         }
+
+        this.columnTypes = this.numerators.reduce((columnTypes, numeratorId, index) => {
```
ethervoid: Why don't we put this inside the async one?

lib/node/nodes/data-observatory-multiple-measures.js
```diff
@@ -49,6 +49,36 @@ var DataObservatoryMultipleMeasures = Node.create(TYPE, PARAMS, { cache: true, l
...
+            };
+            return columnTypes;
+        }, {});
+    },
+    beforeCreateAsync: function(databaseService, callback) {
+        const sql = columnTypesQueryTemplate({
+            source: this.source.getQuery(false),
+            obsMetaParams: this._composeObsMetaParams()
+        });
+        databaseService.run(sql, (err, result) => {
```
ethervoid: Are we going to start using ES6?

lib/node/node.js
rochoa: Because they serve different purposes. The `beforeCreate` validates the input params in an async manner, or let you setup the analysis node from anything that is not async. Look at it almost as a constructor. Some examples:
 - https://github.com/CartoDB/camshaft/blob/5fdd0b57a05a4c71531c300f757eec73783c1c8b/lib/node/nodes/source.js#L10-L19 
- https://github.com/CartoDB/camshaft/blob/5fdd0b57a05a4c71531c300f757eec73783c1c8b/lib/node/nodes/georeference-city.js#L16-L24

The purpose of the new async method is to allow you to run actual async code, as proposed in the PR description:
> Allow executing an async function before the node is created. This allows preparing the node using metadata from the database.

lib/node/nodes/data-observatory-multiple-measures.js
rochoa: I want to default all columns to `::Numeric`, the current behaviour, and later in the async code, only change the columns that I found based on the `numerator_id`, that way I don't have to expect responses in the same order, or metadata for all measures. I want to play safe here.

lib/node/nodes/data-observatory-multiple-measures.js
rochoa: Short answer: YES! 😄 

I started to use it for the templates, where it is quite handy, so I continued the usage for the rest of the code.

lib/node/nodes/data-observatory-multiple-measures.js
ethervoid: Ok!

lib/node/nodes/data-observatory-multiple-measures.js
ethervoid: Good to know! 😄

lib/node/node.js
dgaubert: Hey!

Let me clarify some misconceptions about __releasing Zalgo__.  

### Context
 
We have a class hierarchy between `node` and child implementations such as `GeoreferenceCity` or `source`. 

This PR introduces in `node` the method `beforeCreate` and is intended to be an abstract method with a default implementation (we had it already... implicitly 😄). That method is defined as *sync method* and then every child can implement its own version of `beforeCreate`.

Pretty much the same with `beforeCreateAsync`, here we add a `beforeCreateAsync` abstract method and every child-node can overwrite it. 

The problem here is that the validator calls to `beforeCreate` synchronously during validation. So we don't want to know the details of that method. Then `beforeCreate` must be defined as synchronous method.

On the other hand, `beforeCreateAsync` is called by `analysis-factory` and is intended, as @rochoa said, to fetch some metadata from database. Then `beforeCreateAsync` must be defined as asynchronous method.

If we implement `beforeCreate` expecting a callback and validator never provides it then analysis creation will never end. Viceversa with `beforeCreateAsync`, if we don't provide a callback while calling `beforeCreateAsync` a `undefined is not a function` will be thrown.

### Releasing zalgo

So far, we don't have to worry about `releasing zalgo`. The problem will come if we implements `beforeCreateAsync`, for example, calling the callback asynchronously when everything goes right and calling it synchronously when something goes wrong, usually it appears while checking some condition before calling to the inner async function, i.e:

```
function foo(bar, callback) {
    if (!bar) return callback(new Error('Ooops')); // called synchronously
    baz(bar, function (err, qux) {
         if (err) {
             return callback(err); // called asynchronously
         }
         callback(null, qux); // called asynchronously
    });
}
```

You may notice that we call callbacks synchronously a lot of times in our projects. Actually, it's not so bad thing, only under certain conditions we can block the event loop, such as intensive recursion and so on. Also most of those functions are not public APIs. However, it's a good thing to be aware of it.

Hope this helps.

ethervoid: Understood, thank you for this amazing explanation :)

PR #327, opened 2017-08-29
New limit for input rows multiplied by categories

ethervoid:
Fixes CartoDB/support#895

- Added new limit for input rows multiplied by categories (closest)
- Extract the input/grouped logic to a new function
- Category data now receives an object with source and column nave
- Added new error logs to know when a limit check is failing

### DIFF
--- a/lib/node/limits.js
+++ b/lib/node/limits.js
@@ -7,6 +7,7 @@ var estimates = require('../limits/estimates');
 
 function limitNumberOfRows(node, context, limit, callback) {
     if (!Number.isFinite(limit.value)) {
+        context.logError('Limit for number of rows is not defined');
         return callback(null);
     }
     estimates.estimatedNumberOfRows(node, context, function(err, outputRows) {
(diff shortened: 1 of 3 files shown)


lib/node/limits.js
```diff
@@ -40,12 +42,16 @@ function limitInputRows(node, input, context, limit, callback) {
     });
 }
 
-function limitInputRowsAndAvgGroupedRows(node, input, column, context, lims, callback) {
+function limitInputRowsAndAvgGroupedRows(node, input, categoriesData, context, lims, callback) {
```
rochoa: What about calling this `target` instead of `categoriesData`?

ethervoid: The categories could be in the source (which is how it was before the refactor) so could be a bit missleading

rochoa: What about `categoryTarget`? I don't like the `data` part because it seems like it's already containing the required data.

ethervoid: Hum, sound good to me. Changing it :)

PR #328, opened 2017-08-30
Add ST_MakeValid to the trade-area-dissolved query

ethervoid:
Fixes CartoDB/support#718

Sometimes the isolines generated are invalid geometries and when
we want to dissolve them an exception is raised because ST_Union
doesn't support invalid geometries

Ok, I've done some numbers:

```
30k all invalid:  330707.630 ms
30k with 15% invalid: 120805.738 ms
30k all valid with st_makevalid: 83134.336 ms
30k all valid without makevalid: 64795.720 ms 
---------------------------------------------
5k all invalid: 50399.252 ms
5k with 25% invalid: 13325.140 ms
5k all valid with st_makevalid: 10239.036 ms
5k all valid without makevalid: 8358.520 ms
```

We're are going to introduce a 20% overhead in the dissolve query for trade-areas but I think is the only way. I tried, thanks to the @antoniocarlon's suggestion, to create a function to only apply the `ST_MakeValid` when the `ST_Union` raises an exception but is far worse in performance:

```sql
CREATE OR REPLACE FUNCTION CDB_SafeUnion(geoms geometry[])
RETURNS geometry
AS $$
DECLARE
    geom geometry;
    union_geom geometry;
BEGIN
    FOR geom IN SELECT unnest(geoms)
    LOOP
        BEGIN
            IF union_geom is NULL THEN
                SELECT ST_MakeValid(geom) INTO union_geom;
            ELSE
                SELECT ST_UNION(union_geom, geom) INTO union_geom;
            END IF;
        EXCEPTION WHEN OTHERS THEN
            SELECT ST_UNION(ST_MakeValid(union_geom), ST_MakeValid(geom)) INTO union_geom;
        END;
    END LOOP;
    RETURN union_geom;
END
$$ language plpgsql;
```

So I think the best way is to add `ST_MakeValid` to the dissolve query for the analysis

### DIFF
--- a/lib/node/nodes/sql/trade-area-dissolved.sql
+++ b/lib/node/nodes/sql/trade-area-dissolved.sql
@@ -22,7 +22,7 @@ _cdb_analysis_isochrones_spread AS (
 SELECT
   row_number() over() as cartodb_id,
   data_range,
-  ST_Union(the_geom) the_geom
+  ST_Union(ST_CollectionExtract(ST_MakeValid(the_geom), 3)) the_geom
 FROM
   _cdb_analysis_isochrones_spread
 GROUP BY data_range


rochoa:
Quick questions: why do we have an invalid geometry in the first place? Shouldn't `cdb_dataservices_client.cdb_isochrone` be generating valid geometries?

ethervoid:
Seems to be generating invalid geometries (and invalid geometry could be this kind of polygons:

invalid_polygon

which is possible to have overlap. But I thought like you, that we should not have invalid geometries.

ethervoid:
Could be that we have some kind of issue with the polyline decode [here for HERE](https://github.com/CartoDB/dataservices-api/blob/development/server/lib/python/cartodb_services/cartodb_services/tools/polyline.py) or [here for Mapzen](https://github.com/CartoDB/dataservices-api/blob/development/server/lib/python/cartodb_services/cartodb_services/mapzen/types.py#L31) but at the end is transform a polyline into a polygon using the coordinates provided

ethervoid:
I've been checking what is the root cause of that invalid geometries and it only happens with Mapzen and the is due to the `generalize` option we use to not get huge geometries.

They add a note saying that use their simplification option could lead to a self-intersection (which is the problem we have). You can read it [here](https://mapzen.com/documentation/mobility/isochrone/api-reference/#other-request-parameters)

So we still have to check for the validity of the geometries we have and correct them in case they are wrong but I think is better to include a ST_MakeValid in the Mapzen code instead of the analysis so we have valid geometries.

Closing this PR

Issue #330, opened 2017-09-20
Prevent column name collisions in node queries


jgoizueta:
Some analysis may fail because the produced result tables contain duplicated column names, like in #235 

So we should have a simple way of adapting the name of result columns when needed.

We should preserve the current added column names except when collisions occur, because if
we change a column name in an already cached analysis node, the maps using that analysis will be broken; the list of column names is obtained from the new query (with LIMIT 0) but it is not present in the cached table).

To make things clear for the users we can use the node id (`node.params.id`) when possible as as column name suffix to distinguish from other column of the same name (so we can have names like `count_vals`, `count_vals_a2`)

If we cannot resolve the ambiguity with this method we'll add a numeric suffix, making the number as large as needed (`count_vals`, `count_vals_2`, `count_vals_3`).

jgoizueta:
TODO: we should check all analyses that inherit columns from other nodes and use the ColumnNamer for any columns they add.

rafatower:
@dgaubert I kicked out a duplicated ticket in RT because I thought it can take longer than 3 days, nevertheless I think it's not that hard to fix but important to keep this one in some backlog.

PR #331, opened 2017-09-20
Use unique column names in aggregate-intersection

jgoizueta:
See #330

### DIFF
--- a/lib/node/column-namer.js
+++ b/lib/node/column-namer.js
@@ -7,10 +7,11 @@ function stripTableName(name) {
 // This class is used to help generate the names of node query columns
 module.exports = class ColumnNamer {
     constructor (columns) {
-        this.columns = columns.map(name => stripTableName(name));
+        this.columns = columns.map(name => stripTableName(name).toLowerCase());
     }
 
     uniqueName(baseName) {
+        baseName = baseName.toLowerCase();
         var name = baseName;
         var ok = this.isUnique(name);
         if (!ok) {
@@ -21,12 +22,11 @@ module.exports = class ColumnNamer {
                 ok = this.isUnique(name);
             }
         }
-        this.columns.push(name); // ?
+        this.columns.push(name);
         return name;
     }
 
     isUnique(name) {
-        // TODO: should this be case-insensitive?
         return !this.columns.includes(name);
     }
 
--- a/lib/node/column-namer.js
+++ b/lib/node/column-namer.js
@@ -1,13 +1,13 @@
 'use strict';
 
-function strip_table_name(name) {
+function stripTtableName(name) {
     return name.split('.').pop();
 }
 
 // This class is used to help generate the names of node query columns
 module.exports = class ColumnNamer {
     constructor (columns) {
-        this.columns = columns.map(name => strip_table_name(name));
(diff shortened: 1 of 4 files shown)


jgoizueta:
I'm worried now that maybe using the id from params as a columns name suffix wasn't a good idea: if it is not used to compute the node hash id (is it?) then we could have different queries (differing in result column names) with the same identifier...

lib/node/column-namer.js
```diff
@@ -0,0 +1,33 @@
...
+                ok = this.isUnique(name);
+            }
+        }
+        this.columns.push(name); // ?
+        return name;
+    }
+
+    isUnique(name) {
+        // TODO: should this be case-insensitive?
+        return !this.columns.includes(name);
```
jgoizueta: Can we use  `includes`?

jgoizueta:
Hey @dgaubert, I've added a test to check aggregate-intersection does not produce duplicate column names. It has payed off, because I had missed another potential duplicate columns (`count_vals_density`). I've also checked the test fails because of duplicate `count_vals` with the original aggregate-intersection.

Please take a look when you can.

jgoizueta:
TODO: we should check all analyses that inherit columns from other nodes and use the ColumnNamer for any columns they add.

lib/node/column-namer.js
```diff
@@ -0,0 +1,33 @@
...
+module.exports = class ColumnNamer {
+    constructor (columns) {
+        this.columns = columns.map(name => stripTableName(name).toLowerCase());
+    }
+
+    uniqueName(baseName) {
+        baseName = baseName.toLowerCase();
+        var name = baseName;
+        var ok = this.isUnique(name);
+        if (!ok) {
```
dgaubert: we don't need this initial checking, do we?. I mean if `while(this.isUnique)` is false it skips immediately of 
 suffixing column name, right?

lib/node/nodes/aggregate-intersection.js
```diff
@@ -12,34 +12,46 @@ var PARAMS = {
...
+    var columnNamer = new ColumnNamer(columns);
+
     if (this.aggregate_column === null) {
         qualifiedAggregateColumn += '*';
     } else {
         qualifiedAggregateColumn += this.aggregate_column;
         finalColumnName += '_' + this.aggregate_column;
     }
 
+    if (this.aggregate_function === 'count') {
```
dgaubert: Can we extract everything related to assign a unique name to the output column to a its own method?

lib/node/column-namer.js
jgoizueta: Right! (It was needed before I remove the code to use params.id and didn't realize it wasn't needed anymore)

lib/node/nodes/aggregate-intersection.js
jgoizueta: That line isn't really related to the column renaming, it copes with the different name schema used for `count` aggregations. I moved that bit from where it was to be able to adjust the column name after it was computed.

So we already have all the column-ranaming stuff in this single line:
```
finalColumnName = columnNamer.uniqueName(finalColumnName);
```

PR #346, opened 2017-12-12
Using a partial index over the geometry column of the cached table

### LINKED ISSUE #200: [Analysis] Populate the_geom_webmercator (EPSG:3857) in cache tables
In order to avoid `st_transform(the_geom, 3857)` when rendering.

The best alternative is to use https://github.com/CartoDB/cartodb-postgresql/blob/091aea088e1d37290dbac9d602256d0ececd50bb/scripts-available/CDB_TransformToWebmercator.sql.

Fixes #134.

### LINKED ISSUE #201: [Analysis] Add spatial index on the_geom_webmercator
Probably worthy to explore BRIN as an alternative to GIST index used by CDB_CartodbfyTable.

Depends on #200.

rochoa:
In CARTO, user datasets have their geometries ready for rendering:
they are already projected and have an appropriate index.
On the opposite side, Camshaft's cached tables are not ready for
rendering: missing the adequate projection and the index.

Some numbers on this solution:
https://gist.github.com/rochoa/cf0f512cdad666f62ee5b8930cf4db94.

This is an alternative to and closes #200 and in consequence,
it also closes #201.

### DIFF
--- a/lib/postgresql/node-sql-adapter.js
+++ b/lib/postgresql/node-sql-adapter.js
@@ -40,6 +40,12 @@ NodeSqlAdapter.prototype.checkCacheTableQuery = function() {
     return 'SELECT CDB_CheckAnalysisQuota(\'' + this.node.getTargetTable() + '\')';
 };
 
+NodeSqlAdapter.prototype.createGeomIndexCacheTableQuery = function() {
+    return 'CREATE INDEX ON ' +
+        this.node.getTargetTable() +
+        ' USING GIST (ST_Transform(the_geom, 3857))';
+};
+
 NodeSqlAdapter.prototype.analyzeTableQuery = function() {
     return 'ANALYZE ' + this.node.getTargetTable();
 };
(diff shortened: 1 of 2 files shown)


jorgesancha:
:clap: :clap:

jgoizueta:
The results in the gist look great!

We could look at BRIN indices in the future instead of GIST. They probably won't be a gain in speed up (they're very fast to create but may not optimize the queries too much), but they could be interesting because of the small memory footprint.

I think you already did some (not very successfull) test @rochoa. And we'd need PostgresSQL >= 9.6.

jgoizueta:
By the way, I thought "partial indices" where those with a WHERE condition 🤔

jorgesancha:
@ramiroaznar not a response team ticket but something probably worth including in your bi-weekly update to the rest of the team of work done (with some numbers)

ramiroaznar:
Thanks for the heads up!

PR #349, opened 2018-02-02
add the_geom_webmercator to merge node query

oleurud:
In the JOIN analysis (aka merge node), we are calculating the the_geom_webmercator on the fly. This analysis is not cached, so it is not necessary to calculate it because the source table already has the the_geom_webmercator [more info](https://github.com/CartoDB/camshaft/issues/200#issuecomment-296974271)

This PR adds the `the_geom_webmercator` to the query.

Related with https://github.com/CartoDB/Windshaft-cartodb/pull/862

### DIFF
--- a/lib/node/nodes/merge.js
+++ b/lib/node/nodes/merge.js
@@ -54,10 +54,12 @@ Merge.prototype.sql = function() {
 Merge.prototype._getColumns = function() {
     var geomColumn = ((this.source_geometry === 'left_source') ? INPUT_LEFT_ALIAS : INPUT_RIGHT_ALIAS) +
         '.the_geom as the_geom';
+    var geomWmColumn = ((this.source_geometry === 'left_source') ? INPUT_LEFT_ALIAS : INPUT_RIGHT_ALIAS) +
+        '.the_geom_webmercator as the_geom_webmercator';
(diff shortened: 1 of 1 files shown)


oleurud:
as if it were yours!

lib/node/nodes/merge.js
```diff
@@ -54,10 +54,12 @@ Merge.prototype.sql = function() {
 Merge.prototype._getColumns = function() {
     var geomColumn = ((this.source_geometry === 'left_source') ? INPUT_LEFT_ALIAS : INPUT_RIGHT_ALIAS) +
         '.the_geom as the_geom';
+    var geomWmColumn = ((this.source_geometry === 'left_source') ? INPUT_LEFT_ALIAS : INPUT_RIGHT_ALIAS) +
+        '.the_geom_webmercator as the_geom_webmercator';
```
dgaubert: What happens if one of the sources is another analysis and `the_geom_webmercator` does not exist?

oleurud:
This PR adds a bug: as @jgoizueta said in https://github.com/CartoDB/Windshaft-cartodb/pull/862#pullrequestreview-93955093 , the source layer could be an analysis layer without `the_geom_webmercator`

PR #351, opened 2018-02-07
Change line-sequential legnth units to meters

jgoizueta:
We had this `ST_Length` operating on WGS84 degrees; the rest already do the proper thing operating on geography and returning a meaningful result in meters


dgaubert:
Please add a small note in CHANGELOG.md

PR #353, opened 2018-03-12
TradeArea: Make sure the output 'cartodb_id' is unique

Algunenano:
Related to https://github.com/CartoDB/support/issues/1332

Currently the AOI (non dissolved) analysis returns multiple rows with the same `cartodb_id` (one per isoline) which is troublesome when you do other analyses on top of it. This PR:
* Recalculates cartodb_id to be unique (as with the dissolved analysis).
* Adds a `source_cartodb_id` column in case we want to go back to the original table.
* Changes the `cdb_dataservices_client._cdb_isochrone_exception_safe` fixture to have the same signature as the original function, most importantly returning a set so the issue could be reproduced.

### DIFF
--- a/test/acceptance/trade-area.js
+++ b/test/acceptance/trade-area.js
@@ -46,6 +46,17 @@ describe('trade-area analysis', function() {
                 });
             });
         });
+
+        it('should return unique cartodb_ids', function (done) {
+            testHelper.getResult(tradeAreaDefinition, function(err, result) {
+                assert.ifError(err);
+                result.sort(function(a,b) {
+                    assert.ok(a.cartodb_id !== b.cartodb_id);
+                    return a.cartodb_id < b.cartodb_id;
+                });
+                return done();
+            });
+        });
     });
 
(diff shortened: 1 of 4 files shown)


test/acceptance/trade-area.js
```diff
@@ -46,6 +46,17 @@ describe('trade-area analysis', function() {
                 });
             });
         });
+
+        it('should return unique cartodb_ids', function (done) {
+            testHelper.getResult(tradeAreaDefinition, function(err, result) {
+                assert.ifError(err);
+                result.sort(function(a,b) {
```
dgaubert: what about of using `Set` to get rid of duplicates and compare the lenght of both res and unique ids??:
```
const uniqueIds = [...new Set(result.map(r => r.cartodb_id))]
assert.equal(uniqueCartoDBIds.lenght, result.length)
```

Algunenano: Will do, I didn't know about Set. Thanks!

Issue #268, opened 2017-03-07
No way to apply some analysis more than once


AbelVM:
STR:

You have
* DATASET A : columns `(cartodb_id, the_geom, the_geom_webmercator, kpi1, kpi2)`
* DATASET B and DATASET C: with the same streucture: columns `(cartodb_id, the_geom, the_geom_webmercator, kpi3, a_id)`

You want to join A with B and then with C, so you get a DATASET Z with columns 

(cartodb_id, the_geom, the_geom_webmercator, kpi1, kpi2, **B_kpi3**, **C_kpi3**)

So you perform the first **left join** (left.cartodb_id = right.a_id) and you get

(cartodb_id, the_geom, the_geom_webmercator, kpi1, kpi2, **right_kpi3**)

Now you want to join the third dataset the same way:

YOU GET AN ERROR BECAUSE THERE'S ALREADY A COLUMN CALLED `RIGHT_KPI3`

so you can't perform the join again

cc @dgaubert

ethervoid:
Doing some tests for another issue, I've got into this.

A quick test is to make the source of a trade-area analysis another trade-area analysis. Doing this you're going to get the following error message:

```
Uncaught AssertionError: error: column "center" specified more than once
      + expected - actual

      -false
      +true

      at test/integration/nodes.js:393:24
      at analysis$finish (lib/analysis.js:120:24)
      at node_modules/async/lib/async.js:52:16
      at node_modules/async/lib/async.js:696:30
      at node_modules/async/lib/async.js:167:37
      at node_modules/async/lib/async.js:1209:30
      at lib/workflow/factory.js:108:36
      at lib/service/database.js:146:24
      at lib/service/database.js:112:16
      at lib/postgresql/query-runner.js:37:16
      at Function.<anonymous> (node_modules/cartodb-psql/lib/psql.js:496:17)
      at next [as callback] (node_modules/cartodb-psql/node_modules/step/lib/step.js:51:23)
      at Query.handleError (node_modules/cartodb-psql/node_modules/pg/lib/query.js:129:17)
      at null.<anonymous> (node_modules/cartodb-psql/node_modules/pg/lib/client.js:184:26)
      at Socket.<anonymous> (node_modules/cartodb-psql/node_modules/pg/lib/connection.js:123:12)
      at Socket.<anonymous> (_stream_readable.js:765:14)
      at emitReadable_ (_stream_readable.js:427:10)
      at emitReadable (_stream_readable.js:423:5)
[...]

PR #355, opened 2018-03-13
Chained analyses (Merge and TradeArea)

Algunenano:
While testing the changes I did for https://github.com/CartoDB/support/issues/1298 (avoiding repeated cartodb_id's from analyses) I've faced several times the issue of not being able to create chained analyses from the 2 I was testing (trade area and merge), so here are the fixes for both.

Please let me know if anything needs explanation and I'll try to make it clearer.

I think this also:
* Closes https://github.com/CartoDB/camshaft/issues/268
* References https://github.com/CartoDB/camshaft/issues/330 (Other analyses should still be reviewed)

### DIFF
--- a/lib/node/nodes/merge.js
+++ b/lib/node/nodes/merge.js
@@ -23,9 +23,6 @@ var Merge = Node.create(TYPE, PARAMS);
 
 module.exports = Merge;
 
-var INPUT_LEFT_ALIAS = '_cdb_analysis_left_source';
-var INPUT_RIGHT_ALIAS = '_cdb_analysis_right_source';
-
 var JOIN_OPERATIONS = {
     inner: 'INNER JOIN',
     left: 'LEFT JOIN',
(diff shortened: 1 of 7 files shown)


lib/node/nodes/merge.js
```diff
@@ -23,59 +23,72 @@ var Merge = Node.create(TYPE, PARAMS);
...
     inner: 'INNER JOIN',
     left: 'LEFT JOIN',
     right: 'RIGHT JOIN'
 };
 
 Merge.prototype.sql = function() {
-    var columns = this._getColumns();
+    var buster = util_id.buster();
+
+    var left_alias = '_cdb_analysis_left_source' + buster;
```
dgaubert: 🙏 camelCase

lib/node/nodes/merge.js
```diff
@@ -97,7 +110,7 @@ function setAliasTo(columns, alias, as) {
 
         if (as === 'left') {
             // only set alias to cartodb_id column
-            if (column === 'cartodb_id') {
+            if (/^.*cartodb_id$/.test(column)) {
```
dgaubert: you meant `.includes()` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

lib/node/nodes/trade-area.js
```diff
@@ -18,12 +18,20 @@ var TradeArea = Node.create(TYPE, PARAMS, { cache: true, version: 3, tags: ['io4
 
 module.exports = TradeArea;
 
+const DROPCOLUMN = {
```
dgaubert: `DROP_COLUMN` 😈

lib/node/nodes/merge.js
Algunenano: Not really, I wanted all strings that ended in cartodb_id to catch `cartodb_id` `left_cartodb_id` and `right_cartodb_id`, but now thanks to this I see there is a method called [endsWith](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith) so I'll change it.

Algunenano:
Updated to address comments. Thanks a lot @dgaubert, it seems I end up learning a new JS function with every PR :smile:

lib/node/nodes/merge.js
```diff
@@ -23,59 +23,72 @@ var Merge = Node.create(TYPE, PARAMS);
...
-Merge.prototype._getLeftColumns = function() {
+Merge.prototype._getLeftColumns = function(left_alias) {
     var leftColumns = (this.left_source_columns === null) ?
         this.left_source.getColumns(true) :
         this.left_source_columns;
 
-    leftColumns = setAliasTo(leftColumns, INPUT_LEFT_ALIAS, 'left');
+    if (this.source_geometry === 'left_source') {
+        leftColumns = leftColumns.filter(function (column) {
+            return (/^.*the_geom$/.test(column) === false);
```
enekid: I would extract this function to a `named function` and reuse it in the `_getRightColumns` function:

```javascript
function isNotAGeometryColumn(column) {
    return (/^.*the_geom$/.test(column) === false);
});
```

and then use it like:

```javascript
if (this.source_geometry === 'right_source') {
    rightColumns = rightColumns.filter(isNotAGeometryColumn);
}
```

Algunenano: Nice. I'll also change it to `endsWith` for clarity

PR #362, opened 2018-05-17
Use dollar tokens to scape subqueries received as params for moran analyses

dgaubert:


test/fixtures/cdb_crankshaft/cdb_areasofinterestlocal.sql
```diff
@@ -0,0 +1,23 @@
...
+    quads TEXT,
+    significance NUMERIC,
+    rowid INT,
+    vals NUMERIC)
+AS $$
+BEGIN
+    EXECUTE subquery;
+    SELECT 1, 'QUADS', 0.01, 1, 1;
+END
+$$ LANGUAGE plpgsql VOLATILE PARALLEL UNSAFE;
```
Algunenano: Tests are failing because of the `PARALLEL UNSAFE;`. We strip it in the crankshaft builild script if the DB doesn't support it, but not here. Simply removing it here would be ok.

dgaubert:
That's it, it's an internal change. No need to upgrade anything but camshaft's minor release

PR #366, opened 2018-06-13
Reserved words

enekid:
Quote name columns when performing trade area analysis to avoid Syntax Errors with postgres reserved words.

closes https://github.com/CartoDB/support/issues/1580


lib/node/nodes/trade-area.js
```diff
@@ -30,9 +30,9 @@ TradeArea.prototype.sql = function() {
     var template = this.dissolved ? tradeAreasDissolvedQueryTemplate : tradeAreasQueryTemplate;
     const columnsQuery = this.source.getColumns(true)
         .filter(columnName => !(DROP_COLUMN[columnName]))
-        /*jshint unused: false */
-        .map(columnName => `'{$columnName}'`)
-        /*jshint unused: true */
+        /*jshint quotmark: false */
+        .map(columnName => "\"" + columnName + "\"")
```
dgaubert: what abaut \`"{$columnName}"\` ?

enekid: It didn't work because I don't know who write string templates properly:
``` 
`"{$columnName}"` 
````

vs

``` 
`"${columnName}"` 
````

File: test/acceptance/trade-area.js [hunk 2]
```diff
@@ -57,6 +57,34 @@ describe('trade-area analysis', function() {
             });
         });
 
+        it('should not fail with Postgres reserved words', function (done) {
```
jgoizueta: we could also add test for column names that contain dashes, spaces, now that we support them

PR #367, opened 2018-06-15
Geocoder boost

juanignaciosl:
CartoDB/product/issues/92


lib/node/nodes/georeference-street-address.js
```diff
@@ -30,6 +30,7 @@ GeoreferenceStreetAddress.prototype.sql = function() {
     var sql = queryTemplate({
         source: this.source.getQuery(),
         columns: this.source.getColumns(true).join(', '),
+        street_column: this.getStreetAddressColumnParam(),
```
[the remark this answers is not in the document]
juanignaciosl: In fact I still have to dig into the details of how columns are composed :+1:

juanignaciosl:
@rochoa could you CR both this and CartoDB/dataservices-api/pull/499 , please ? I'd like you to review both because the dataservices one contains a lot of ~hacks~ dynamic query execution and you might propose some improvements. No hurry, this won't be merged until all providers are ready, and I can begin working on that.

lib/node/nodes/georeference-street-address.js
```diff
@@ -49,42 +53,55 @@ GeoreferenceStreetAddress.prototype.getGeocoderParams = function () {
...
 GeoreferenceStreetAddress.prototype.getStreetAddressColumnParam = function() {
     if (this.street_address_column) {
         return this.getFunctionParam('street_address');
     }
 
-    var columns = [];
-    var result = this.street_address_template.replace(COLUMNS_TEMPLATE_REGEX, function(s, column) {
-        columns.push(column.trim());
-        return '%s';
+    var columns_and_literals = this.street_address_template.trim().split(', ').map(function(fragment) {
```
rochoa: This is a bit complex to understand.

Why do we split by `', '`?

lib/node/nodes/georeference-street-address.js
```diff
@@ -49,42 +53,55 @@ GeoreferenceStreetAddress.prototype.getGeocoderParams = function () {
...
+        fragment = fragment.trim();
+        if(COLUMNS_TEMPLATE_REGEX.test(fragment)) {
+            var column_match = COLUMNS_TEMPLATE_REGEX.exec(fragment);
+            return column_match[1].trim();
+        } else {
+            return '\'\'' + fragment + '\'\'';
```
rochoa: At the time we worked in this node, we couldn't use ES6 string interpolation/template literals. I encourage you to take advantage of them for better readability.

lib/node/nodes/georeference-street-address.js
```diff
@@ -49,42 +53,55 @@ GeoreferenceStreetAddress.prototype.getGeocoderParams = function () {
...
     });
 
-    if (columns.length === 0) {
-        return '$tpl$' + this.street_address_template + '$tpl$';
+    if (columns_and_literals.length === 0) {
+        return '$tpl$\'' + this.street_address_template + '\'$tpl$';
     }
 
-    return 'format($tpl$'+result+'$tpl$, ' + columns.join(', ') + ')';
+    return '$tpl$' + columns_and_literals.join(' || \'\', \'\' || ') + '$tpl$';
```
rochoa: This produces outputs like `$tpl$''Logroño'' || '', '' || ''La Rioja'' || '', '' || ''Spain''$tpl$`. Is that intentional for some reason I'm not understading in this context?

juanignaciosl: Yes. it's used in the template, at `street_column`:

https://github.com/CartoDB/camshaft/pull/367/files#diff-6a7ad66626fc4d94638530f35f0ba1abR9. In that usage it's not SQL code yet: it's a string that will be executed by the extension. That string should be a valid SQL fragment.

lib/node/nodes/georeference-street-address.js
rochoa: I guess tests were "_happy_" use-cases, but the previous code didn't assume any format as it can be any free text coming from the user. For instance, a user could input `{{street_number}} {{street_name}}. {{postal_code}} {{city}}`.

juanignaciosl: Ok, I'll improve that, taking into account that I can't just split by spaces in order to support things like `{{street_number}} {{street_name}}, La Rioja`.

juanignaciosl:
@rochoa what about another CR? :pray: 

PS: I think we'll soon forget those dark JS days without string interpolation :D.

lib/node/nodes/georeference-street-address.js
```diff
@@ -49,42 +53,58 @@ GeoreferenceStreetAddress.prototype.getGeocoderParams = function () {
...
+            columnsAndLiterals.push(column.trim());
+        } else {
+            columnsAndLiterals.push(`'${addressMatch[0]}'`);
+        }
+    }
 
-    if (columns.length === 0) {
-        return '$tpl$' + this.street_address_template + '$tpl$';
+    if (columnsAndLiterals.length === 0) {
+        return `$tpl$'${this.street_address_template}'$tpl$`;
```
rochoa: Just wondering. Why did we have to abandon the `format` approach? I guess we cannot pass that to the remote server to be lazily evaluated but we can do it with a string. 🤔

lib/node/nodes/georeference-street-address.js
```diff
@@ -49,42 +53,58 @@ GeoreferenceStreetAddress.prototype.getGeocoderParams = function () {
...
 };
 
 var queryTemplate = Node.template([
     'SELECT',
-    '  {{=it.columns}},',
-    '  cdb_dataservices_client._cdb_geocode_street_point_exception_safe(' +
-    '    {{=it.params}}',
-    '  ) AS the_geom',
-    'FROM ({{=it.source}}) AS _camshaft_georeference_street_address_analysis'
- ].join('\n'));
```
rochoa: What about taking advantage of ES6 templates/interpolation also here?

test/acceptance/georeference-street-address.js
```diff
@@ -50,75 +48,143 @@ describe('georeference-street-address analysis', function() {
...
+            query: 'select * from georeference_street_address_fixture where cartodb_id between 1 and 3',
+            addresses: [
+                { point: { x: -61.1, y: -29.1 } },
+                { point: { x: -61.2, y: -29.2 } },
+                { point: { x: -61.3, y: -29.3 } }
+            ]
+        },
+        {
+            desc: 'template with two columns without commas',
+            template: '{{street_name}} {{street_number}}, {{city}}, Spain',
```
rochoa: 👍

lib/node/nodes/georeference-street-address.js
juanignaciosl: The reason was the lazy evaluation. Previously, extracting the columns was enough. Now, to enable lazy evaluation, literals must be escaped.

I don't know what you mean by "doing it with a string", could you elaborate, please? TBH, this was one of the first changes, maybe it's not needed. Take a look at dataservices side as well. This kind of design decisions are the ones that made me ask for your opinion on this :-)

rochoa: Sorry about my wording. By "doing it with a string", I meant something similar to what you say with "literals must be escaped". I tend to use `format` over string concatenation, but I guess it's not an option here.

juanignaciosl: Yep, I think that it's not possible (or makes things more complicated, I'm not sure).

rochoa: Let's :shipit: like this.

antoniocarlon:
HERE tested in staging. Everything is working as expected :tada:

antoniocarlon:
If any of the components is `None` [here](https://github.com/CartoDB/dataservices-api/pull/499/files#diff-425264fb6b9c402688b61f31580a92d8R96), this will raise `'NoneType' object has no attribute 'encode'`.
Can be fixed doing `s.encode("utf-8") if s != None else s`

juanignaciosl:
@antoniocarlon :+1: at acceptance as well for this one, right?

antoniocarlon:
Yes! :rocket:

test/acceptance/georeference-street-address.js
```diff
@@ -50,75 +48,143 @@ describe('georeference-street-address analysis', function() {
...
-            y: 42.302939
+            query: 'select * from georeference_street_address_fixture where cartodb_id = 1',
+            addresses: [
+                { cartodb_id: 1, point: { x: -2.517555, y: 42.302939 }, street_name: 'W 26th Street' }
+            ]
+        },
+        {
+            desc: 'multiple rows',
+            column: 'full_address',
+            query: 'select * from georeference_street_full_address_fixture',
```
dgaubert: Please use the clause `ORDER BY` for this test. This test against `< Postgres@10` will fail because the default order has changed for `Postgres@10` as the assertion phase will loop over the result in the very same order that it was queried.

CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 # Changelog
+## 0.62.0
+Release 2018-mm-dd
+  - Add support for batch street-level geocoding analysis.
+
 ## 0.61.12
 Release 2018-mm-dd
```
dgaubert: Hey `0.61.12` was not released, the last one was `0.61.11` which is the one with release date defined. You need to update this section (the title changes to 0.62.0).

Issue #364, opened 2018-05-28
Add .npmignore


Algunenano:
```
npm notice === Tarball Details === 
npm notice name:          camshaft                                
npm notice version:       0.61.10                                 
npm notice package size:  139.7 kB                                
npm notice unpacked size: 1.3 MB                                  
npm notice shasum:        9055bbc577dbd87d38a0e712bf6157afd7704dca
npm notice integrity:     sha512-O07OZQsx1ULNY[...]1IV//IHwZWWDg==
npm notice total files:   243                                     
npm notice 
```

As with the other projects, we are publishing much more than what is really needed.

cc/ @CartoDB/rt-managers

rafatower:
```
$ npm pack camshaft
camshaft-0.61.11.tgz

$ tar -xzf camshaft-0.61.11.tgz

$ ncdu package

  968.0 KiB [##########] /reference                                                                                                                                                                                                                                                       
  528.0 KiB [#####     ] /test
  448.0 KiB [####      ] /lib
  136.0 KiB [#         ] /examples
   20.0 KiB [          ] /docs
   20.0 KiB [          ] /config
   20.0 KiB [          ] /tools
   16.0 KiB [          ]  CHANGELOG.md
    8.0 KiB [          ]  .jshintrc
    8.0 KiB [          ] /.github
    4.0 KiB [          ]  LICENSE
    4.0 KiB [          ]  HOWTORELEASE.md
    4.0 KiB [          ]  .travis.yml
    4.0 KiB [          ]  package.json
    4.0 KiB [          ]  Makefile
    4.0 KiB [          ]  README.md
    4.0 KiB [          ]  CONTRIBUTING.md
    4.0 KiB [          ]  .npmignore
```

The `reference/` dir is actually its own package and could be `.npmignore`d. In turn `camshaft-reference` should have its own `.npmignore`, I think.

PR #368, opened 2018-07-06
Add .npmignore #364

rafatower:
This closes https://github.com/CartoDB/camshaft/issues/364

This is pretty similar to what's done in https://github.com/CartoDB/Windshaft/pull/636

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -2,7 +2,6 @@
 # https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package
 
 # Biggest dirs not needed in npm package
-reference/
 test/
 examples/
 
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,13 @@
+# Please read carefully before editing, and especially removing lines:
+# https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package
+
+# Biggest dirs not needed in npm package
+reference/
+test/
+examples/
+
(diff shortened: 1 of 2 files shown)


rafatower:
I lost a lot of time after fixing the issue mentioned above, because of yarn. I was basically testing the same old broken package again and again :cry: 

In the end, this:
```
yarn cache clean
```

made the trick.

Now it works :smile: : 
```
$ rm -rf node_modules && yarn install && yarn test && echo "ALL GOOD"

[...]

  multilayer:postgres=layergroup_url=/api/v1/map
Dependency="camshaft" installed version="0.61.11" does not match declared version="file:/tmp/test-camshaft-package/camshaft-0.61.11.tgz". Check your installation.
    ✓ layergroup with 2 layers, each with its style (190ms)
    ✓ get creation requests has cache (66ms)
    ✓ get creation has no cache if sql is bogus
    ✓ get creation has no cache if cartocss is not valid
    ✓ layergroup can hold substitution tokens (146ms)
    ✓ layergroup creation raises mapviews counter (49ms)
    ✓ layergroup creation fails if CartoCSS is bogus
    ✓ layergroup creation fails if sql is bogus
    ✓ layergroup with 2 private-table layers (152ms)

[...]

  1042 passing (1m)
  17 pending

[...]

***jshint***
Done in 104.43s.
ALL GOOD
```

PR #370, opened 2018-07-25
Upgraded Street analysis version

juanignaciosl:
- [X] If it modifies the output columns, updates the analysis version
- [X] Has tests to validate the result is the expected one.


juanignaciosl:
This was fixed at Camshaft level, no need to upgrade version if you add new columns and you don't want analysis to be run again.

Issue #361, opened 2018-03-16
Map instantiation: Get columns from cache table instead of using `LIMIT 0` when avaiblable


Algunenano:
After the recent issues with analyses changing column names, I think it'd better that during map instantiation instead of doing `_cdb_analysis_query LIMIT 0` to know the columns of the analysis we checked the cached table columns instead when available.

Although the version update would still be required this would benefit us 2 ways:
* Maps won't be broken after an update.
* Analysis update won't be mandatory (since it doesn't break the map) so we could just alert the user and let it be run by them at their convenience (or not be rerun at all).

This is somewhat of a small project so not suited for RT. @rochoa can you please have a look and see if this fits in the Product roadmap or could be moved to any squad (backend or core would be my guess) .

cc/ @dgaubert

rochoa:
I don't understand the context/problem. 😕

Algunenano:
Case 1:
* Modify a analysis (for example AOI) with cache and change the name of an output column.
* Forget to update version.
Current result: Map fails to instantiate and it's inaccessible in builder.

Case 2:
* Modify a analysis (for example AOI) with cache and change the name of an output column.
* Update version.
Current result: The analysis is rerun automagically.

This would address the first case and enables that the rerun in Case 2 is optional (since the map wouldn't be broken).

rochoa:
Understood. Thank you.

About the "Case 2": As re-running is optional, how do we enable the user to re-run it?

Algunenano:
My idea is that this could be done in the UI after the instantiation (something like `This analysis can be updated`).
That should be another ticket after this one since it'd involve:
* Not doing the version update automatically when a difference is found.
* Returning extra information in the instantiation.
* Adding the UI stuff with the alert to trigger the update.

rochoa:
And how that would work at API level?

Algunenano:
My initial idea is to add an extra (non-persisted) state to the analysis, `upgradable` which means that the analysis is `ready` but there is an update available. Then to update it you could push the new config as if the parameters had changed.

The analysis backend to handle changes seems to be defined in Builder (`app/controllers/carto/api/analyses_controller.rb` and `app/models/carto/analysis.rb`) so I expect some changes to be needed there to handle version mismatches in a different way.

juanignaciosl:
We've not deployed the new metadata columns for street geocoding because this affects us. Old analyses fail with this error:

[...]

PR #372, opened 2018-07-31
Get query from cache to compute output columns when building analysis

dgaubert:
Tested in https://github.com/CartoDB/Windshaft-cartodb/pull/1015

Related to #361


juanignaciosl:
:heart:

PR #378, opened 2018-12-17
Change error message for large payload errors

jgoizueta:
The batch SQL API returns a 400 when the queries payload is over 16KB. In this case the generic message about not being able to queue the job is not very useful.


jgoizueta:
Can we merge this @dgaubert?

dgaubert:
Yes

PR #379, opened 2019-02-12
Improve Batch API error message

rafatower:
Follow up of this PR: https://github.com/CartoDB/camshaft/pull/378

This improves the error message for exceeded batch SQL API payload size by adding suggestions about what the user can do about it.


rafatower:
Adding you to review mostly for awareness. Aside, I'm afraid of being a bit rusted with camshaft deploys :man_shrugging:

PR #385, opened 2019-06-27
Do not get affected tables and the last updated time 

dgaubert:
when a foreign table is involved


rafatower:
Related to https://github.com/CartoDB/cartodb-platform/issues/5755

Issue #397, opened 2020-02-17
Uncaught exception: Range filter expect to have at least one value in greater_than, greater_than_or_equal, less_than, less_than_or_equal, min, or max numeric params


jgoizueta:
Moved from https://github.com/CartoDB/Windshaft-cartodb/issues/1117

An exception (`Error: Range filter expect to have at least one value in greater_than, greater_than_or_equal, less_than, less_than_or_equal, min, or max numeric params`) can occur while executing `sqlWrappedNode.createTableQuery` and should be catched and returned via the `callback`.

PR #398, opened 2020-02-17
Catch invalid range exception

jgoizueta:
See #397

### DIFF
--- a/lib/service/database.js
+++ b/lib/service/database.js
@@ -110,14 +110,18 @@ DatabaseService.prototype.createCacheTable = function(node, callback) {
             }
 
             var sqlWrappedNode = new NodeSqlAdapter(node);
-            self.queryRunner.run(sqlWrappedNode.createTableQuery(), QUERY_RUNNER_WRITE_OP, function(err) {
-                debug('TODO: register table');
-                if (err && !DUPLICATED_TABLE_ERROR_CODES[err.code]) {
-                    return callback(err, tableExists);
-                }
+            try {
+                self.queryRunner.run(sqlWrappedNode.createTableQuery(), QUERY_RUNNER_WRITE_OP, function(err) {
+                    debug('TODO: register table');
+                    if (err && !DUPLICATED_TABLE_ERROR_CODES[err.code]) {
+                        return callback(err, tableExists);
+                    }
 
-                return callback(null, tableExists);
-            });
+                    return callback(null, tableExists);
+                });
+            } catch (err) {
+                return callback(err);
+            }
         });
     } else {
         return callback(null, false);
--- a/lib/service/database.js
+++ b/lib/service/database.js
@@ -108,20 +108,21 @@ DatabaseService.prototype.createCacheTable = function(node, callback) {
             if (tableExists) {
                 return callback(null, tableExists);
             }
-
             var sqlWrappedNode = new NodeSqlAdapter(node);
+            let createTableQuery;
             try {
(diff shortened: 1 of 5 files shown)


test/integration/factory.js
```diff
@@ -51,4 +56,56 @@ describe('factory', function() {
...
         });
     });
+
+    it('invalid filters in createCacheTable', function(done) {
+        const TEST_SOURCE_TYPE = 'test-source';
+        const TestSource = Node.create(TEST_SOURCE_TYPE, {
+            table: Node.PARAM.STRING()
+        }, { cache: true });
+        TestSource.prototype.sql = function() {
+            return 'select * from ' + this.table;
```
dgaubert:
```suggestion
            return `select * from ${this.table}`;
```

test/integration/factory.js
```diff
@@ -51,4 +56,56 @@ describe('factory', function() {
...
+                source: sourceDefinition,
+                sampling: 1,
+                seed: 12345,
+                filters: invalid_filters
+            }
+        };
+        const definition = {
+            type: 'sampling',
+            params: {
+                source: samplingDefinition,
```
dgaubert: Do we need to nest two sampling nodes?

test/integration/factory.js
```diff
@@ -51,4 +56,56 @@ describe('factory', function() {
             return done();
         });
     });
+
+    it('invalid filters in createCacheTable', function(done) {
+        const TEST_SOURCE_TYPE = 'test-source';
+        const TestSource = Node.create(TEST_SOURCE_TYPE, {
+            table: Node.PARAM.STRING()
+        }, { cache: true });
+        TestSource.prototype.sql = function() {
```
dgaubert: Long time I worked on this library, so I might forget a lot of context. So my following question might not make sense: why do we need a `test-source` type?

jgoizueta: Good question, I'm not sure either, I picked it from the existing test in that file, it's a cached source, it worked for me so I used it (but I guess I could use an actual Source).

test/integration/factory.js
jgoizueta: Yes, I could only reproduce the exception within `createTableQuery` that way. With just one the exception occurs after createTableQuery (in getColumns) where it is already handled correctly.

test/integration/factory.js
```diff
@@ -51,4 +57,56 @@ describe('factory', function() {
...
+            return 'select * from ' + this.table;
+        };
+
+        const invalid_filters = {
```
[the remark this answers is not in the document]
jgoizueta: oops

lib/service/database.js
```diff
@@ -110,14 +110,18 @@ DatabaseService.prototype.createCacheTable = function(node, callback) {
...
-            self.queryRunner.run(sqlWrappedNode.createTableQuery(), QUERY_RUNNER_WRITE_OP, function(err) {
-                debug('TODO: register table');
-                if (err && !DUPLICATED_TABLE_ERROR_CODES[err.code]) {
-                    return callback(err, tableExists);
-                }
-
-                return callback(null, tableExists);
-            });
+            try {
+                self.queryRunner.run(sqlWrappedNode.createTableQuery(), QUERY_RUNNER_WRITE_OP, function(err) {
```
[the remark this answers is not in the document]
jgoizueta: you're right there, we need to catch just around createTableQuery

PR #403, opened 2020-03-26
Create geomval in cdb_dataservices_client schema

Algunenano:
Minor changes in the tests to create the geomval type, if necessary, in the cdb_dataservices_client schema. This is done to match the behaviour of that extension.

No need to deploy, this only changes tests and I did it to make sure that we are testing the same behaviour as the one that, in the end, was implemented in the dataservices extension.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## 0.65.4
 Release 2020-mm-dd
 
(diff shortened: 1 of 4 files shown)


Algunenano:
Checking the failures. They are related to the search_path (which cartodb users have, but not newly created ones). I'll look for a way to implement it in the tests.

manmorjim:
nice! :rocket:

PR #407, opened 2020-08-25
Stop logging node input params

dgaubert:
cc @jvillarf


oleurud (approved):

jvillarf:
Thanks a lot. Let's check it today in staging and double check that there are no more UUIDs indexed :hugs:

PR #409, opened 2020-10-29
[ch89482] Error with column names with capital letters

Shylpx:
### Resources

- [Clubhouse story](https://app.clubhouse.io/cartoteam/story/89482/mr-jeff-if-a-column-name-has-an-uppercase-it-throws-an-error-when-run-add-column-from-second-dataset)
- [PostgreSQL - SQL syntax identifiers](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS)

### Context

- There is an error when an analysis is performed using columns with capital letters. This is due to, according to the [PostgreSQL doc](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS), the identifiers are always lower-cased in order to perform the query. And that is the reason why, when the column name is "Example", it raises an error saying there is no column called "example".
- It is not the reported issue, but if the column name has spaces, the problem would be very similar.
- The standard behaviour is not to have column with capital letters or spaces, but the column names can be changed for example using an `ALTER TABLE ... RENAME COLUMN ... TO ...`.

### Changes

- Use double quotes around identifiers to force use the column names as they have been written.


dgaubert (changes requested):
I'm missing a test to validate the new behavior. As a reference, take any test that uses `type: 'merge'` in https://github.com/CartoDB/camshaft/blob/master/test/integration/nodes.js, for instance: https://github.com/CartoDB/camshaft/blob/master/test/integration/nodes.js#L519-L545

HTH

Shylpx:
I have added the test for the new behavior. Can you review it again? Thanks, @dgaubert !

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.67.2
+Release 2020-10-29
```
dgaubert, line 4: Update de date please!

dgaubert (approved):
Impressive work! The way you implemented the test, forcing to fail the test first, and then make it pass.... pretty neat!

Update the CHANGELOG, then merge it. Let me know if you need anything for the release!

PR #416, opened 2021-12-27
[sc138209] Fix column names for Merge analyses

Shylpx:
### Resources

- [Shortcut story](https://app.shortcut.com/cartoteam/story/138209/elmundodata-click-popup-not-showing-info)

### Context

- In https://github.com/CartoDB/camshaft/pull/355/, was added support for `Merge` chained analyses: `_cdb_analysis_left_source`/`_cdb_analysis_right_source` was being used as query alias. This is ok if there is only one `Merge` analysis, but if you add two analyses of this type to the same layer, a name conflict appears in the SQL query. To solve that, a `buster` was added at the end of every alias, just a numeric code generated using the current timestamp.
- This timestamp worked to allow chained analyses, but this causes the query to change every time the map is instantiated, generating cache issues, as described [here](https://app.shortcut.com/cartoteam/story/138209/elmundodata-click-popup-not-showing-info#activity-141539). So, we need each analysis with unique alias names, but this code shouldn't change every time the map is rendered.

### Changes

- Use analysis node ID as `buster` instead of current time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.67.4
+Release 2021-12-28
+
+- Fixes:
+  - Use node ID as buster for alias names on Merge analyses to avoid cache issues
+
 ## 0.67.3
 Release 2021-02-15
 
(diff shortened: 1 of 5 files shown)


moicalcob (approved):
LGTM! 😄

